PR #24, opened 2023-08-21
wr-11: Protected Routing | FE

hnoievets:
Was implemented route wrapper component that redirects not logged-in users from protected routes to Sign In page.


frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
+import { Navigate } from 'react-router-dom';
+
+import { AppRoute } from '~/libs/enums/app-route.enum';
+import { useAppSelector } from '~/libs/hooks/hooks.js';
+
+type Properties = {
+  component: React.FC
+};
+
+const ProtectedRoute: React.FC<Properties> = ({ component: Component, ...rest }) => {
```
andanf-e: Should not take `component` prop - should wrap routes from `children`
to use it like this:
```
<ProtectedRoute>
  <Page />
</ProtectedRoute>
```

```suggestion
const ProtectedRoute: React.FC<Properties> = ({ children }) => {
```

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const { user }  = useAppSelector( ({ auth }) => ({
+    user: auth.user
+  }) );
+
+  const hasUser = Boolean(user);
+
+  return hasUser ? (
+    <Component {...rest} />
+  ) : (
+    <Navigate to={{ pathname: AppRoute.SIGN_IN }} />
```
andanf-e: Also add optional prop `redirectPath` with default value `AppRoute.SIGN_IN`
```suggestion
    <Navigate to={redirectPath} replace />
```

andanf-e: Let's better use early return for such cases - it may be more readable
```suggestion
  if (!hasUser) {
    return <Navigate ..>;
  }
  
  return <>{children}</>;
```

frontend/src/packages/auth/auth.ts
```diff
@@ -11,3 +11,4 @@ const authApi = new AuthApi({
 });
 
 export { authApi };
+export { type UserAuthResponseDto } from './libs/types/types.js';
```
NazarKovtun1: Should be moved to a special reexport file (src/libs/types/user/user.ts), then reexport it from src/libs/types/user.ts in src/libs/types/types.ts and import it always from there. All these reexports are required foe clean importing routes

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
+import { Navigate } from 'react-router-dom';
```
andanf-e: And add reexport of it on `libs/components/components`
```suggestion
import { Navigate } from '~/libs/components/components.js';
```

frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { AppRoute } from '~/libs/enums/app-route.enum';
```
andanf-e:
```suggestion
import { AppRoute } from '~/libs/enums/enums.js';
```

andanf-e (changes requested):
And you've missed EOL on each file
Do you use editorconfig extension? If not - it's helpful

frontend/src/packages/auth/auth.ts
NazarKovtun1: Oh, also, reexport it in frontend/src/packages/auth/libs/types and use wherever you need. Do not export type from the place it should not be exported (frontend/src/packages/auth/auth.ts)
@andanf-e right?

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -2,15 +2,18 @@ import { createSlice } from '@reduxjs/toolkit';
 
 import { DataStatus } from '~/libs/enums/enums.js';
 import { type ValueOf } from '~/libs/types/types.js';
+import { type UserAuthResponseDto as User } from '~/packages/auth/auth.js';
```
NazarKovtun1: import this type from packages/auth/libs

NazarKovtun1 (changes requested):

frontend/src/packages/auth/auth.ts
andanf-e: Actually it may be reexported from domain root reexport file, but in our case it should be not `auth` but `users` package - so on `shared` folder it should be declared in `packages/users/libs/types` and in `frontend` - should be reexported also from `users`

mhrechkin (dismissed):

frontend/src/packages/auth/libs/types/types.ts
```diff
@@ -0,0 +1 @@
+export { type UserAuthResponseDto } from 'shared/build/index.js';
```
andanf-e: As it is now reexported through `package/users`
```suggestion

```

andanf-e (changes requested):

frontend/src/packages/auth/libs/types/types.ts
hnoievets: It was added after the merge(

andanf-e (approved):

NazarKovtun1 (approved):

mhrechkin (approved):

PR #28, opened 2023-08-22
wr-15: create header component

PablloZz:


frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,27 @@
+import avatar from '~/assets/img/user-avatar.svg';
+import { Link } from '~/libs/components/link/link.jsx';
+
+import styles from './styles.module.scss';
+
+type Properties = {
+  user: unknown;
```
andanf-e, line 7: And after auth changes implemented we'd change unknown to actual type
```suggestion
  user: unknown | null;
```

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,27 @@
...
+const Header: React.FC<Properties> = ({ user }) => (
+  <>
+    {user && (
+      <header className={styles.header}>
+        <Link to="/">
+          <span className={styles.logo}>WRITORIUM</span>
+        </Link>
+        <div className={styles['avatar-info']}>
+          <Link to="/profile">
+            <img src={avatar} alt="avatar" className={styles.avatar} />
```
andanf-e, line 19: Let's create separate common component for avatar - as we may use it in different places throughout the app
it may take user name initials (first to letters of first/last name -- we will use it as placeholder when there is no avatar image applied) and `avatarUrl: string | null` 
In case if user has avatar uploaded - then we'd show it as image, in other case we'd show placeholder with initials
something like this
image

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type Properties = {
+  user: unknown;
+};
+
+const Header: React.FC<Properties> = ({ user }) => (
+  <>
+    {user && (
+      <header className={styles.header}>
+        <Link to="/">
```
andanf-e:
```suggestion
        <Link to={AppRoute.ROOT}>
```

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,27 @@
...
+          <span className={styles.logo}>WRITORIUM</span>
+        </Link>
+        <div className={styles['avatar-info']}>
+          <Link to="/profile">
```
andanf-e:
```suggestion
          <Link to={AppRoute.PROFILE}>
```

andanf-e (changes requested):

frontend/src/libs/components/app/app.tsx
```diff
@@ -27,6 +29,7 @@ const App: React.FC = () => {
 
   return (
     <>
+      <Header user={null} />
```
WilhelmWesser, line 32: When auth is implemented, do not forget to change this passed value user from the store

NazarKovtun1, line 32: When auth is implemented, do not forget to change this passed value user from the store

NazarKovtun1 (dismissed):

frontend/src/libs/components/header/header.tsx
PablloZz: Can I leave as it is for now? 
Because there is an error from eslint:
type Properties = {
  user: unknown | null;
};
'unknown' overrides all other types in this union type.

frontend/src/libs/components/header/header.tsx
PablloZz: Do the colors have to be the same for a non-existent avatar or can I use green for example?

mhrechkin (dismissed):

frontend/src/libs/components/header/header.tsx
PablloZz: On daily you said that there's layout for non-existing avatar but there's no it yet, so can I leave for now something like this?
image

PablloZz: So while the default user avatar layout isn't done, can I use something like this for now?

.default-avatar {
  display: flex;
  justify-content: center;
  align-items: center;
  font-size: 20px;
  font-weight: bold;
  background: lightblue;
  color: rgb(17, 159, 206);
}

andanf-e: A little bit corrected:
```
/** @markup */
<div className={styles.defaultAvatar}>
  <span className={styles.initials}>{initials}</span>
</div>

/** @styling */
 /** colors and vars for them taken from Andrew Bobrovsky's message in slack */
.defaultAvatar {
  display: flex;
  justify-content: center;
  align-items: center;
  background-color: #44996B; // --light-green
}
.initials {
  color: #2E453B; // --dark-green
  font-size: 20px;
  font-weight: 500;
  text-transform: uppercase;
}
```

PR #34, opened 2023-08-22
wr-5: Sign In | BE

AnastasiaR2:
1.  Model getting user for sign-in checkin
2.  Service sign-in user flow
3.  Controller sign-in route
4.  Validation shared with FE


backend/src/libs/interfaces/repository.interface.ts
```diff
@@ -1,5 +1,6 @@
 interface IRepository<T = unknown> {
   find(): Promise<T>;
+  findByEmail(email: string): Promise<T | null>;
```
andanf-e: It should not be a part of all repositories - it should be implemented only upon user repository
```suggestion
```

backend/src/libs/interfaces/service.interface.ts
```diff
@@ -1,5 +1,6 @@
 interface IService<T = unknown> {
   find(): Promise<T>;
+  findByEmail(email: string): Promise<T | null>;
```
andanf-e: It should not be a part of all services - it should be implemented only upon user service
```suggestion
```

shared/src/libs/exceptions/not-found-error/not-found-error.exception.ts
```diff
@@ -0,0 +1,14 @@
+type Constructor = {
+  message: string;
+  cause?: unknown;
+};
+
+class NotFoundError extends Error {
```
andanf-e: This error is only a part of backend exceptions - so it should be declared only on backend

backend/src/packages/auth/auth.service.ts
```diff
@@ -11,6 +15,35 @@ class AuthService {
...
+    const mockedUserResponse: UserAuthResponseDto = {
+      id: userRequestDto.id,
+      email: userRequestDto.email,
+      firstName: 'Test',
+      lastName: 'User',
+    };
+    return Promise.resolve({
+      ...mockedUserResponse,
+    });
+  }
```
andanf-e:
```suggestion
  public async signIn(
    userSignInDto: UserSignInResponseDto,
  ):Promise<UserAuthResponseDto> {
    const { email } = userSignInDto;
  
    const user = await this.userService.findByEmail(email);
    
    if (!user) {
      /** message and cause should be predefined as default values for argument on exception */
      throw new NotFoundError();
    }

    /** firstName|lastName may be mocked up on entity level, because after implementation they may be fulfilled there */
    return user;
  }
```

backend/src/packages/auth/auth.controller.ts
```diff
@@ -64,7 +78,47 @@ class AuthController extends Controller {
...
+  private async signIn(
+    options: ApiHandlerOptions<{
+      body: UserAuthRequestDto;
+    }>,
+  ): Promise<ApiHandlerResponse> {
+    const user = await this.authService.verifySignInCredentials(options.body);
+    return {
+      status: HttpCode.OK,
+      payload: await this.authService.signIn(user),
+    };
```
andanf-e:
```suggestion
    return {
      status: HttpCode.OK,
      payload: await this.authService.signIn(options.body),
    };
```

backend/src/packages/auth/auth.controller.ts
```diff
@@ -6,8 +6,8 @@ import {
 } from '~/libs/packages/controller/controller.js';
 import { HttpCode } from '~/libs/packages/http/http.js';
 import { type ILogger } from '~/libs/packages/logger/logger.js';
-import { type UserSignUpRequestDto } from '~/packages/users/users.js';
-import { userSignUpValidationSchema } from '~/packages/users/users.js';
+import { type UserAuthRequestDto, type UserSignUpRequestDto } from '~/packages/users/users.js';
+import { userSignInValidationSchema, userSignUpValidationSchema } from '~/packages/users/users.js';
```
andanf-e: These imports may be joined
```suggestion
import {
  type UserAuthRequestDto,
  type UserSignUpRequestDto,
  userSignInValidationSchema,
  userSignUpValidationSchema
  } from '~/packages/users/users.js';
```

andanf-e (changes requested):

backend/src/packages/auth/auth.controller.ts
NazarKovtun1: Also, to simplify, you can try following syntax - import type { UserAuthRequestDto, UserSignUpRequestDto,  userSignInValidationSchema, userSignUpValidationSchema } from '~/packages/users/users.js';
As far, as I remember, modern TS supports that, @andanf-e right?

andanf-e: I think it (`import type { UserAuthRequestDto, UserSignUpRequestDto, userSignInValidationSchema, userSignUpValidationSchema } from '~/packages/users/users.js';`) won't work because like this you're declaring all import parts as types -- so this way schemas would be marked for TS also as types and it would cause errors

the main goal with such kind of import
```
import {
  type UserAuthRequestDto,
  type UserSignUpRequestDto,
  userSignInValidationSchema,
  userSignUpValidationSchema
}
```
with `type ` beginings for type imports is to tell TS earlier about things that may be skipped when it transpiles code to js

backend/src/packages/auth/auth.service.ts
```diff
@@ -11,6 +15,32 @@ class AuthService {
...
+  ): Promise<UserSignInResponseDto> {
+    const user = await this.userService.findByEmail(userRequestDto.email);
+
+    if (!user) {
+      throw new UserNotFoundError();
+    }
+
+    return user;
+  }
+
```
andanf-e:
```suggestion

```

andanf-e (changes requested):

mhrechkin (dismissed):

andanf-e (approved):

NazarKovtun1 (approved):

PR #39, opened 2023-08-22
wr-19: Loader Component

hnoievets:
Added Loader component that:

-  wraps component and shows loader upon `isLoading` prop;
- gets an optional `className` prop.

image

### DIFF
--- a/frontend/src/libs/components/loader/enum/default-color.enum.ts
+++ b/frontend/src/libs/components/loader/enum/default-color.enum.ts
@@ -1,0 +1,6 @@
+const DefaultLoaderColor = {
+  BORDER: '#2e453b',
+  BORDER_TOP: 'transparent',
+};
+
+export { DefaultLoaderColor };
--- a/frontend/src/libs/components/loader/enum/default-color.enum.ts
+++ b/frontend/src/libs/components/loader/enum/default-color.enum.ts
@@ -1,6 +1,0 @@
-const DefaultLoaderColor = {
-  BORDER: '#2e453b',
-  BORDER_TOP: 'transparent',
-};
-
-export { DefaultLoaderColor };
--- a/frontend/src/libs/components/loader/loader.tsx
+++ b/frontend/src/libs/components/loader/loader.tsx
@@ -1,0 +1,32 @@
+import { DefaultLoaderColor } from './enum/default-color.enum.js';
+import styles from './styles.module.scss';
+
+type Properties = {
+  children: React.ReactNode;
+  color?: string;
+  isLoading: boolean;
+};
+
+const Loader: React.FC<Properties> = ({
+  children,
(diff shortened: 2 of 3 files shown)


frontend/src/libs/components/loader/enum/default-color.enum.ts
andanf-e, line 1: You don't need it

frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  return (
+    <div className={styles.container}>
+      <div
+        className={styles.loader}
+        style={{
+          borderColor: color,
+          borderTopColor: DefaultLoaderColor.BORDER_TOP,
+        }}
```
andanf-e:
```suggestion

```

frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,32 @@
...
+  isLoading,
+}) => {
+  if (!isLoading) {
+    return <>{children}</>;
+  }
+
+  return (
+    <div className={styles.container}>
+      <div
+        className={styles.loader}
```
andanf-e: Better provide default color in styles and if we'd need to provide custom styles - then it would be provided by className from optional prop
```suggestion
        className={getValidClassNames(styles.loader, className)}
```

frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,32 @@
+import { DefaultLoaderColor } from './enum/default-color.enum.js';
+import styles from './styles.module.scss';
+
+type Properties = {
+  children: React.ReactNode;
+  color?: string;
```
andanf-e:
```suggestion
  className?: string;
```

frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,32 @@
...
+  isLoading: boolean;
+};
+
+const Loader: React.FC<Properties> = ({
+  children,
+  color = DefaultLoaderColor.BORDER,
```
andanf-e:
```suggestion
  className,
```

andanf-e (changes requested):

frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,32 @@
...
+  isLoading,
+}) => {
+  if (!isLoading) {
+    return <>{children}</>;
```
artemmatiushenko1: Looks like fragment is unnecessary here. We can simply write `return children`

mhrechkin (dismissed):

andanf-e (approved):

NazarKovtun1 (approved):

PR #45, opened 2023-08-23
wr-8: Sign Up | FE

artemmatiushenko1:
Implemented common layout for authorization pages. 

To display background letters I used just a text with a corresponding font. The reason I chose this aprroach is because the approach which is based on exporting all letters as transparent image and embedding it as backgound image doesn't produce the expected result and looks far too different from the design.


frontend/src/pages/auth/auth.tsx
```diff
@@ -44,10 +44,14 @@ const Auth: React.FC = () => {
...
-      {getScreen(pathname)}
-    </>
+    <AuthLayout
+      formElement={
+        <>
+          state: {dataStatus}
+          {getScreen(pathname)}
+        </>
+      }
+    />
```
andanf-e: Let's pass form element as children
```suggestion
    <AuthLayout>
      <>
        state: {dataStatus}
        {getScreen(pathname)}
      </>
    </AuthLayout>
```

frontend/src/pages/auth/components/auth-layout/auth-layout.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type Properties = {
+  formElement: ReactNode;
+};
+
+const AuthLayout: React.FC<Properties> = ({ formElement }) => {
+  const renderFloatingLetters = (letters: FloatingLetter[]): ReactNode[] =>
+    letters.map((item) => (
+      <div
+        key={item.value}
```
andanf-e: Using item value (which is letter), may lead to some unexpected behavior - as these letters may be the same (like duplicating W and R) and duplicate keys for elements may cause something bad with rendering

frontend/src/pages/auth/components/auth-layout/libs/types/floating-letter.type.ts
```diff
@@ -0,0 +1,11 @@
+type FloatingLetter = {
+  value: string;
+  position: {
+    left?: number;
+    top?: number;
+    bottom?: number;
+    right?: number;
```
andanf-e: I believe they are not all optional,, but could exist in tuples, like
```
type Position = Record<'left' | 'top', number> | Record<'left' | 'bottom', number> | Record<'right' | 'top', number> | Record<'right'|'bottom', number>
```

andanf-e (changes requested):

mhrechkin (dismissed):

frontend/src/pages/auth/components/auth-layout/auth-layout.tsx
```diff
@@ -0,0 +1,40 @@
...
+type Properties = {
+  children: ReactNode;
+};
+
+const AuthLayout: React.FC<Properties> = ({ children }) => {
+  const renderFloatingLetters = (letters: FloatingLetter[]): ReactNode[] =>
+    letters.map((item) => (
+      <div
+        key={item.key}
+        style={{ ...item.position }}
```
NazarKovtun1: redundant destructurization as you pass nothing else there
```suggestion
        style={item.position}
```

NazarKovtun1 (changes requested):

andanf-e (dismissed):
LGTM, waiting for rest changes

mhrechkin (dismissed):

NazarKovtun1 (dismissed):

drMangust:
I believe it would be beneficial for us to explore the possibility of establishing a unified style for "background letters." This would enable us to ensure consistency and avoid the need for different implementations in various locations. Looking ahead, I'd like to suggest that openly discussing thoughts and ideas regarding specific implementations could lead to greater productivity. It appears that we have previously discussed these matters, but somehow a different decision was reached. Upon reflection, had an alternative implementation not been agreed upon, it could have resulted in your time being spent without purpose. I greatly value your innovative solution. However, considering the collaborative nature of our team, enhancing communication among us could prove even more invaluable.

frontend/src/pages/auth/components/auth-layout/auth-layout.tsx
```diff
@@ -0,0 +1,43 @@
+import { type ReactNode } from 'react';
+
+import {
+  AUTH_FORM_WRAPPER_FLOATING_LETTERS,
+  HERO_FLOATING_LETTERS,
+} from './libs/constants.js';
+import { type FloatingLetter } from './libs/types/floating-letter.type.js';
+import styles from './styles.module.scss';
```
PablloZz, line 8: I think it should be exported from `'./libs/types/types.js';``

frontend/src/pages/auth/components/auth-layout/auth-layout.tsx
```diff
@@ -0,0 +1,43 @@
+import { type ReactNode } from 'react';
+
+import {
+  AUTH_FORM_WRAPPER_FLOATING_LETTERS,
+  HERO_FLOATING_LETTERS,
+} from './libs/constants.js';
+import { type FloatingLetter } from './libs/types/floating-letter.type.js';
```
PablloZz: I think it should be imported from `'./libs/types/types.js';``

andanf-e (dismissed):

b0bbobovich:
After successfull registration you must add user token to localstorage. Add this logic to auth.actions.ts file.

b0bbobovich:
Also you need to modify validation scheme for sign-up in the shared folder to satisfy follow criterias: 
Untitled

andanf-e:
> Also you need to modify validation scheme for sign-up in the shared folder to satisfy follow criterias: Untitled

Let's merge basic functionality and extend it + setting user token to localStorage is a part of Auth Token ticket

andanf-e (approved):

NazarKovtun1 (approved):

PR #46, opened 2023-08-23
wr-41: Create icon component

PablloZz:


frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,13 @@
+import { type IconName, iconNameToIcon } from './common.js';
+
+type Properties = {
+  iconName: IconName;
```
andanf-e, line 4: Let's also add optional `className` prop for icons
```suggestion
  className?: string;
  iconName: IconName;
```

andanf-e (changes requested):

andanf-e (approved):

mhrechkin (approved):

PR #55, opened 2023-08-24
wr-21: Add select component

### LINKED ISSUE #21: Select Component
image

- [x] Add wrapper component for https://react-select.com/home
- [x] Provide default styling for this component from design
- [x] Provide ability to pass custom styling
- [x] Should be controlled by react-hook-forms

**User story:**
As a user, I want to be able to use a Select Component, So that I can choose one option from a list of options.

**Acceptance criteria:**

- [ ] A dropdown-style list displayed with a default placeholder "Search...".
- [ ] When a user clicks on the Select Component a list of selectable options should be displayed.
- [ ] When a user clicks on one of the options the selected option should be visually highlighted and the dropdown list should collapse, displaying the selected option as the current value.
- [ ] When no option is selected, field should display a placeholder.
- [ ] When a user clicks outside the Select Component the dropdown list should close, hiding the options.
- [ ] If there is an issue with loading options for the Select Component an error message should be displayed to indicate the issue.

canterbery:
1.  Add wrapper component for https://react-select.com/home
2.  Provide default styling for this component from design
3.  Provide ability to pass custom styling
4.  Should be controlled by react-hook-forms


frontend/src/libs/components/icon/common.ts
```diff
@@ -1,3 +1,4 @@
+import { ReactComponent as Arrow } from '~/assets/img/icons/arrow.svg';
```
andanf-e: I believe `ArrowDown` is the same thing - if so then we don't need a duplicate

frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,110 @@
...
+  control,
+  errors,
+  label,
+  ...restProperties
+}: Properties<T, IsMulti, Group>): JSX.Element => {
+  const { field } = useFormController({ name, control });
+  const error = errors[name]?.message;
+  const hasError = Boolean(error);
+
+  const defaultStyles: StylesConfig<SelectOption, IsMulti, Group> = {
```
andanf-e: Let's move defaultStyles to `./libs/constants` for this component

frontend/src/libs/types/select-option.type.ts
```diff
@@ -0,0 +1,6 @@
+type SelectOption = {
+  value: string;
```
andanf-e: value may also be number, not only string

andanf-e (changes requested):

frontend/src/libs/components/select/select.tsx
canterbery: I tried to do it but there are type issues if I move it outside.  I can pass SelectOption and Group outside but not isMulti.  
I tried to skip it but it's a second variable so something like StyleConfig<SelectOption, Group> doesnt work because it expects second type variable to extend boolean (i.e to be isMulti)

andanf-e: Maybe then we can wrap `defaultStyles` into `getDefaultStyles` helper - then method could have generic typing that would be identical to generic passed for Select component

frontend/src/libs/components/icon/common.ts
canterbery: Not exactly. My svg contains only arrow but arrowDown contains arrow and white space around it. And if I use it I get either too small arrow or wrong height of select component
I guess I can replace arrowDown with my arrow then.

andanf-e: Yep 👍

NazarKovtun1 (dismissed):

andanf-e (approved):

mhrechkin (approved):

PR #56, opened 2023-08-24
wr-43: Create common Modal Component

hnoievets:
Added a `Modal ` component that may take children to be displayed. Used a `Portal` for Modal implementation.

Знімок екрана 2023-08-24 174616
Знімок екрана 2023-08-24 180313


frontend/src/libs/components/modal/styles.module.scss
```diff
@@ -0,0 +1,61 @@
...
+  padding: 5px 5px 0 0;
+  background-color: var(--white);
+  border: none;
+  border-radius: 5px;
+  cursor: pointer;
+}
+
+@media only screen and (width < 768px) {
+  .content {
+    width: 95%;
```
andanf-e, line 35: Restrict max-width, instead of width

frontend/src/libs/components/modal/styles.module.scss
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+@media only screen and (width >= 1200px) {
+  .content {
+    width: 900px;
+  }
+}
+
+@media only screen and (width >= 1920px) {
```
andanf-e: declare screen breakpoints in `variables`

andanf-e (changes requested):

NazarKovtun1 (dismissed):

frontend/src/libs/hooks/use-modal/use-modal.hook.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type ReturnValue = {
+  isOpen: boolean;
+  toggle: () => void;
+};
+
+const useModal = (): ReturnValue => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  const toggle = (): void => {
```
andanf-e:
```suggestion
  const handleToggleModalOpen = (): void => {
```

andanf-e (changes requested):

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  onClose: () => void;
+  isOpen: boolean;
+};
+
+const Modal: React.FC<Properties> = ({
+  isOpen,
+  onClose: handleOnClose,
+  children,
+}) => {
+  const outsideModalDivReference = createRef<HTMLDivElement>();
```
artemmatiushenko1: Please use `useRef` hook instead, `createRef` is usually used in class components. I guess `overlayRef` would be a better variable name

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type Properties = {
+  children: React.ReactNode;
+  onClose: () => void;
+  isOpen: boolean;
+};
+
+const Modal: React.FC<Properties> = ({
+  isOpen,
+  onClose: handleOnClose,
```
artemmatiushenko1: I guess there's no need to rename the prop

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+      if (event.target !== outsideModalDivReference.current) {
+        return;
+      }
+      handleOnClose();
+    },
+    [handleOnClose, outsideModalDivReference],
+  );
+
+  useEffect(() => {
+    const outsideModalDiv = outsideModalDivReference.current;
```
artemmatiushenko1: `const overlayElement = `

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return (
+    <Portal>
+      <div className={styles.modal} ref={outsideModalDivReference}>
+        <div className={styles.content} role="button" tabIndex={0}>
+          <button className={styles.closeBtn} onClick={handleOnClose}>
+            <Icon iconName="crossMark" />
+          </button>
+
```
artemmatiushenko1, line 48: Unnecessary empty line

frontend/src/libs/components/modal/modal.tsx
hnoievets: Thanks

frontend/src/libs/components/modal/modal.tsx
hnoievets: Handlers located in the component must have the `handle` prefix.

artemmatiushenko1: Hmm, but actually it's a prop, not a handler. Maybe I'm wrong 😅 @andanf-e could you please suggest here

frontend/src/libs/components/modal/modal.tsx
hnoievets: I did this because when I showed this code to others, they defined the parent div for content as just a button wrapper.

frontend/src/libs/components/modal/modal.tsx
andanf-e: @artemmatiushenko1 you are right
Yep, it's a handler passed as a prop, so it s hould be named as it's passed -- actually may reason with such kind of naming convention is to differentiate own component handlers (`handleKek`) and passed methods (`onKek`)

hnoievets: > Handlers located in the component must have the `handle` prefix.

This is a quote from Mykola when I used `onClose` as an `onClick` hanler. OK, I will change it again)

hnoievets: Thanks

frontend/src/libs/hooks/hooks.ts
```diff
@@ -1,7 +1,9 @@
 export { useAppDispatch } from './use-app-dispatch/use-app-dispatch.hook.js';
 export { useAppForm } from './use-app-form/use-app-form.hook.js';
 export { useAppSelector } from './use-app-selector/use-app-selector.hook.js';
+export { useModal } from './use-modal/use-modal.hook.js';
 export { useMousePosition } from './use-mouse-position/use-mouse-position.hook.js';
 export { useCallback, useEffect, useMemo, useState } from 'react';
+export { useRef } from 'react';
```
andanf-e: join it with other hook imports from react

andanf-e (changes requested):

frontend/src/libs/components/portal/styles.module.scss
```diff
@@ -0,0 +1,12 @@
+.noOverflow {
+  overflow: hidden;
+}
+
+.portal {
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 1000;
```
andanf-e: Move `z-index` also to variables, and I am not sure that we need `1000`

andanf-e (changes requested):

frontend/src/libs/components/portal/styles.module.scss
hnoievets: What value do we need?

hnoievets: Is 100 enough?

andanf-e: I believe yes

andanf-e: You may name it something like `z-index-level-one`

hnoievets: Got it. Thanks

andanf-e (approved):

NazarKovtun1 (approved):

frontend/src/libs/components/portal/portal.tsx
```diff
@@ -0,0 +1,37 @@
...
+const Portal: React.FC<Properties> = ({ children }) => {
+  const portalContainer = useMemo(() => {
+    const element = document.createElement('div');
+    element.classList.add(styles.portal);
+
+    return element;
+  }, []);
+
+  useEffect(() => {
+    const wasOverflowHidden = document.body.classList.contains(
```
nickolay-grechkin, line 20:
```suggestion
    const isOverflowHidden = document.body.classList.contains(
```

PR #59, opened 2023-08-24
wr-31: Implement article saving logic | BE

### LINKED ISSUE #31: Implement article saving logic | BE
**User story:**
As a writer, I want to have control over publishing and drafting articles, so that I can decide whether an article should be published immediately or saved as a draft, and it appears correctly in the “My Articles” tab and feed.

**Acceptance criteria:**
- [x] After clicking the “Publish” button, the article should be listed in the “My Articles” tab on the “My articles” page. The article should also appear in the feed for all users to see.
- [x] The articles table migration  for creating table should include the “published_at” column, for indicating the publication date of articles on the feed.
- [x] After clicking the “Save Draft” button, the article should be listed in the “My Articles” tab  on the “My articles” page with a “draft” label.

image

drMangust:
- [x] articles migration
- [x] implement article entity
- [x] implement article model
- [x] implement article repository
- [x] implement article service
- [x] Implement validation schemas
- [x] implement article controller

In conclusion: **implementation of the article saving logic was made**

Notes: 
- At the moment the userId is mocked in the code. We need an authorizationPlugin that will add the user to the request

image
image

Postman tests: 
image
image
image
image
image
image


backend/src/db/migrations/20230822132525_create_articles_table.ts
```diff
@@ -0,0 +1,48 @@
+import { type Knex } from 'knex';
+
+const TableName = {
```
mhrechkin, line 3: I think that you don't need this enum as you can use DatabaseTableName enum which is defined in backend/src/libs/packages/database/libs/enums/database-table-name.enum.ts

backend/src/packages/articles/article.repository.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  public async findById(id: number): Promise<ArticleEntity | null> {
+    const article = await this.articleModel.query().findById(id).execute();
+
+    if (!article) {
+      return null;
+    }
+
+    return ArticleEntity.initialize(article);
+  }
```
andanf-e: It may be done in find (there is changes in Authorization plugin PR) - so don't forget to move it to find after that changes merged

backend/src/packages/articles/article.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  public async findById(id: number): Promise<ArticleGetItemResponseDto | null> {
+    const article = await this.articleRepository.findById(id);
+
+    if (!article) {
+      return null;
+    }
+
+    return article.toObject();
+  }
```
andanf-e: Same as for comment in repository

backend/src/packages/articles/article.repository.ts
drMangust: @andanf-e Hi. Thanks for the review.
- I didn't do it in find, only because I wanted to have a simpler answer from the method. because in the future we will be returning full article data, with user data, comments, etc. So I decided that I didn't need that much in the context of what I did it for.

- I didn't quite understand about the Authorization plugin PR, what it has to do with find. or I'm missing the context. I have certain things to do when the Authorization plugin is ready, but they have to do with getting the userId in the controller

andanf-e: I've checked how you'd use it - so currently your purpose is to check that article actually exists (for that purpose we may create `checkIfExists` in repository and it would look like (`select 1 from articles where id = $id` in raw), because later on we would actually have `findById` with whole relations and it would confuse + find method in abstract repository was actually added to be extended into `find (id: number)`

drMangust: Hmmm..., a little bit wrong, I use findById for:
1. Yes, to check if the record exists in the database.
2. the received data is only the fields from this table I take as default, for transmission with payload

image

mhrechkin (changes requested):

backend/src/db/migrations/20230822132525_create_articles_table.ts
andanf-e: Yep, we can,, and for database enum it may not be something bad (cause we won't want to change any database tables names),, but overall it's better to don't use anything from outer code in migrations

mhrechkin: Okay, we can left it as it is)

mhrechkin (dismissed):

backend/src/packages/articles/article.repository.ts
andanf-e: Okay, then still lets `find(id:number)` would be default findById,, and when we'd need to find with relations - create method for it separately 🤔

drMangust: - findById methods became find
- merged the branch with the "Authorization Plugin". And checked that everything was working as it should. ⬇️

Screenshot 2023-08-25 173637

Screenshot 2023-08-25 174352

image

```ts
[17:36:17.524] INFO (12220): POST on /api/v1/auth/sign-up
[17:43:04.418] INFO (12220): POST on /api/v1/articles/
[17:45:29.146] INFO (12220): PUT on /api/v1/articles/5
```

backend/src/packages/articles/article.service.ts
drMangust: findById method became find

backend/src/db/migrations/20230822132525_create_articles_table.ts
drMangust: Ok.

backend/src/packages/articles/article.entity.ts
```diff
@@ -0,0 +1,96 @@
...
+    return new ArticleEntity({
+      id,
+      title,
+      text,
+      userId,
+      promptId,
+      genreId,
+      publishedAt,
+    });
+  }
```
NazarKovtun1, line 51: I think, we can simplify this as this destructurization just duplicates fields
```suggestion
  public static initialize(articleEntity: ArticleEntityType): ArticleEntity {
    return new ArticleEntity(articleEntity);
  }
```

NazarKovtun1: Also, check other places in your code where this suggestion is applicable. @andanf-e @nickolay-grechkin what to you think?

shared/src/packages/articles/libs/enums/article-validation-message.enum.ts
```diff
@@ -0,0 +1,14 @@
+const ArticleValidationMessage = {
+  ARTICLE_TITLE_REQUIRE: 'Title is required',
+  ARTICLE_TITLE_MIN_LENGTH: 'Title must contain at least 3 characters',
+  ARTICLE_TITLE_MAX_LENGTH: 'Title should contain no more than 30 characters',
```
andanf-e: Use `ArticleValidationRule` also for validation message

```suggestion
  ARTICLE_TITLE_MAX_LENGTH: `Title should contain no more than ${ArticleValidationRule.TITLE_MAX_LENGTH} characters`,
```

andanf-e (changes requested):

backend/src/packages/articles/article.entity.ts
drMangust: I think I can do that now. And when ArticleEntity expands, we may think about what to change.

drMangust: When I was doing it, I was looking at UserEntity as an example because it was in the starter

drMangust: Well, the main idea is that the ArticleEntityType itself can be wider, but I only need these fields in the output

andanf-e: Yep @drMangust you're right - because we may have wider data shape - we may form it only with needed things

shared/src/packages/articles/libs/enums/article-validation-message.enum.ts
drMangust: Done - [76a0c38](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/59/commits/76a0c382b4836eca13173448752ca51314fe61f3)

backend/src/packages/articles/article.entity.ts
drMangust: @andanf-e So do we leave it as it is?

andanf-e (approved):

backend/src/packages/articles/article.entity.ts
NazarKovtun1: Oh, Okay then, agreed

NazarKovtun1 (approved):

PR #62, opened 2023-08-25
wr-30: implement save article form

OlenaMohylivska29:


drMangust:
Are you sure you need to merge into my "task/wr-31-articles" branch?

andanf-e:
> Are you sure you need to merge into my "task/wr-31-articles" branch?

I am sure that I want to see only this branch's diff - so it is a common way to open PRs based not on the default env branch -- to point it on branch that is was checkouted from,, and then - when base branch was merged, repoint it to needed one

drMangust:
> 

I have not yet had to do this. Interesting.

drMangust:
> > Are you sure you need to merge into my "task/wr-31-articles" branch?
> 
> I am sure that I want to see only this branch's diff - so it is a common way to open PRs based not on the default env branch -- to point it on branch that is was checkouted from,, and then - when base branch was merged, repoint it to needed one

I have not yet had to do this. Interesting.

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
...
+    (buttonName: string) =>
+      (payload: ArticleCreateRequestDto): void => {
+        const updatedPayload = {
+          ...payload,
+          ...(buttonName === 'publish' && {
+            publishedAt: new Date().toISOString(),
+          }),
+        };
+
+        void dispatch(articlesActions.createArticle(updatedPayload));
```
andanf-e, line 35: Split it into 2 different actions - `publishArcticle` and `saveArticleDraft`,, and let's not prepopulate publishedAt from FE

andanf-e (changes requested):

frontend/src/libs/components/article-form/article-form.tsx
drMangust: If I don't pass publishedAt from the FE, then how do I distinguish whether it's a save or a publication? My logic is as follows: saving and publication follow the same route, if the field is in the request (I expect the date in ISO 8601 format from the FE), then if nothing interferes (validation of other fields), then it writes the data to the publishedAt field.

drMangust: @andanf-e 
Maybe we misunderstood?

Once again, the way I see it, and by the way, if I remember correctly, it was originally described in the task

1. when creating a publication, the publishedAt field is optional. if it is, then it is a publication. If not, it is a save.
2. When updating, I also provided the ability to pass the publishedAt field as optional, and I think it can be passed to null if, for example, the user wants to remove an article from published.

Or there is another option, to make this logic, all through the same routes, but to pass some options through query parameters.

@OlenaMohylivska29, @andanf-e, @nickolay-grechkin, @WilhelmWesser Let's discuss it, share your vision and ideas?

OlenaMohylivska29: One more option is to pass the extra header  XActionType: 'publish', but I am inclined to use query params too

andanf-e: Yep, sorry
I've overthought it a little bit -- let's keep it as it is with one route and publishedAt payload

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { actions as articlesActions } from '~/slices/articles/articles.js';
+
+import { Button, Input, Textarea } from '../components.js';
+import styles from './styles.module.scss';
+
+const DEFAULT_ARTICLE_FORM_PAYLOAD: ArticleCreateRequestDto = {
+  title: '',
+  text: '',
+  genreId: 1,
+};
```
andanf-e: Move it to `./libs/constants`

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+  const handleArticleSubmit = useCallback(
+    (buttonName: string) =>
+      (payload: ArticleCreateRequestDto): void => {
+        const updatedPayload = {
+          ...payload,
+          ...(buttonName === 'publish' && {
+            publishedAt: new Date().toISOString(),
+          }),
```
andanf-e: Let's keep `publishedAt` field consistent in payload

```suggestion
          publishedAt: buttonName === 'publish' ? new Date().toIsoString() : null
```

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
...
+const ArticleForm: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { control, errors, handleSubmit, reset } =
+    useAppForm<ArticleCreateRequestDto>({
+      defaultValues: DEFAULT_ARTICLE_FORM_PAYLOAD,
+      validationSchema: articleCreateValidationSchema,
+    });
+
+  const handleArticleSubmit = useCallback(
+    (buttonName: string) =>
```
andanf-e:
```suggestion
    (submitType: 'draft' | 'publish') =>
```

NazarKovtun1 (dismissed):

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,97 @@
...
+const ArticleForm: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { control, errors, handleSubmit, reset } =
+    useAppForm<ArticleCreateRequestDto>({
+      defaultValues: DEFAULT_ARTICLE_FORM_PAYLOAD,
+      validationSchema: articleCreateValidationSchema,
+    });
+
+  const handleArticleSubmit = useCallback(
+    (submitType: 'draft' | 'publish') =>
```
mhrechkin: Move it to enum

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,97 @@
...
+        void dispatch(articlesActions.createArticle(updatedPayload));
+      },
+    [dispatch],
+  );
+
+  const handleFormSubmit = useCallback(
+    (event_: React.BaseSyntheticEvent<SubmitEvent>): void => {
+      const button = event_.nativeEvent.submitter as HTMLButtonElement;
+      void handleSubmit(
+        handleArticleSubmit(button.name as 'draft' | 'publish'),
```
mhrechkin: Move it to enum

mhrechkin (changes requested):

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
+import { ButtonType } from '~/libs/enums/button-type.enum.js';
+import { SubmitType } from '~/libs/enums/enums.js';
```
andanf-e:
```suggestion
import { ButtonType, SubmitType } from '~/libs/enums/enums.js';
```

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const dispatch = useAppDispatch();
+  const { control, errors, handleSubmit, reset, dirtyFields } =
+    useAppForm<ArticleRequestDto>({
+      defaultValues: DEFAULT_ARTICLE_FORM_PAYLOAD,
+      validationSchema: articleCreateValidationSchema,
+    });
+
+  const areFieldsDirty = !!dirtyFields.title && !!dirtyFields.text;
+
+  type SubmitType = ValueOf<typeof SubmitType>;
```
andanf-e: 1 - Don't place type declarations inside components
2 - Use type in `ValueOf<typeof SubmitType>` form, because now it looks confusing that here we can find SubmitType as enum and SubmitType as type
```suggestion

```

frontend/src/libs/components/article-form/article-form.tsx
andanf-e, line 1: I guess it should be placed not in common components - but let's resolve this later (when we'd have articles page)
For now leave a @todo comment - that it should be moved out from common components

frontend/src/libs/enums/components/article-form/article-form-submit-type.enum.ts
```diff
@@ -0,0 +1,6 @@
+const SubmitType = {
```
andanf-e: And name enum the same way + not sure if it should be in common libs/enums - as it is related only to articles form
```suggestion
const ArticleSubmitType = {
```

frontend/src/pages/auth/components/sign-in-form/sign-in-form.tsx
```diff
@@ -60,6 +60,7 @@ const SignInForm: React.FC<Properties> = ({ onSubmit }) => {
           Sign Up
         </Link>
       </div>
+      <ArticleForm />
```
andanf-e: Sign In form is not a place where we want it to be placed
```suggestion

```

frontend/src/slices/articles/articles.slice.ts
```diff
@@ -0,0 +1,37 @@
...
+const { reducer, actions, name } = createSlice({
+  initialState,
+  name: 'articles',
+  reducers: {},
+  extraReducers(builder) {
+    builder.addCase(createArticle.pending, (state) => {
+      state.dataStatus = DataStatus.PENDING;
+    });
+    builder.addCase(createArticle.fulfilled, (state, action) => {
+      state.articles.push(action.payload);
```
andanf-e: I guess redux-toolkit's immer will deal with it, but let's make it not in that rough way
```suggestion
      state.articles = state.articles.concat(action.payload);
```

andanf-e (changes requested):

frontend/src/libs/components/article-form/article-form.tsx
AnastasiaR2: Will it be correct, if I rename the type to ArticleSubmitTypeValue and move it to the types folder for article-form?

andanf-e: Just use enum with `ValueOf<typeof [enum]>`

frontend/src/libs/components/article-form/article-form.tsx
```diff
@@ -0,0 +1,106 @@
...
+// TODO: Should be moved out from common components.
+const ArticleForm: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { control, errors, handleSubmit, reset, dirtyFields } =
+    useAppForm<ArticleRequestDto>({
+      defaultValues: DEFAULT_ARTICLE_FORM_PAYLOAD,
+      validationSchema: articleCreateValidationSchema,
+    });
+
+  const areFieldsDirty = !!dirtyFields.title && !!dirtyFields.text;
```
andanf-e: Is there any considerations not to use isDirty from useForm instead of manually check each dirtyField?

AnastasiaR2: I tried to use `disabled={!isDirty}` on the “Save Draft” and “Publish” buttons, but in this case, they become available if any of the fields are not empty. But they should be enabled only if both title and text are not empty. Maybe there is an approach with isDirty I missed?

drMangust: isDirty is for whether the fields have changed. And I used it to disable the save button in edit user info

andanf-e: Let's handle it with isDirty per whole form, and mark this discussion in comment to ticket (to make it clear with acceptance criterias)

AnastasiaR2: For now, I just need to write `disabled={!isDirty || isSubmitting}` in both buttons?

andanf-e: Yep

andanf-e (dismissed):

andanf-e (approved):

mhrechkin (approved):

PR #64, opened 2023-08-26
wr-57: Create common ErrorMessage Component

AnastasiaR2:
1. Component takes validation error prop and optional prop for custom styling. 
2. It has default styling that is relative to form field error styling. 
3. Applied to input and select components.


frontend/src/libs/components/select/select.tsx
```diff
@@ -62,7 +63,7 @@ const Select = <
         styles={{ ...getDefaultStyles<IsMulti, Group>(), ...styles }}
         components={{ DropdownIndicator, IndicatorSeparator }}
       />
-      {hasError && <span className={cssStyles.error}>{error as string}</span>}
+      {hasError && <ErrorMessage error={error as string} />}
```
andanf-e:
```suggestion
      <ErrorMessage error={error as string} />
```

frontend/src/libs/components/input/styles.module.scss
```diff
@@ -20,7 +21,7 @@
 }
 
 .error {
-  margin-bottom: 5px;
+  margin-bottom: 0;
```
andanf-e:
```suggestion

```

frontend/src/libs/components/input/styles.module.scss
```diff
@@ -2,6 +2,7 @@
   width: 100%;
   height: 45px;
   margin-top: 5px;
+  margin-bottom: 22px;
```
andanf-e: I believe you don't need this
```suggestion

```

frontend/src/libs/components/input/input.tsx
```diff
@@ -53,11 +54,7 @@ const Input = <T extends FieldValues>({
         type={type}
         placeholder={placeholder}
       />
-      {hasError && (
-        <span className={getValidClassNames(styles.text)}>
-          {error as string}
-        </span>
-      )}
+      {hasError && <ErrorMessage error={error as string} />}
```
andanf-e:
```suggestion
      <ErrorMessage error={error as string} />
```

frontend/src/libs/components/error-message/error-message.tsx
```diff
@@ -0,0 +1,18 @@
...
+  error: string;
+  className?: string;
+};
+
+const ErrorMessage: React.FC<Properties> = ({ error, className }) => {
+  return (
+    <div className={getValidClassNames(styles.errorMessage, className)}>
+      {error}
+    </div>
+  );
```
andanf-e:
```suggestion
  const hasError = Boolean(error);

  return (
    <span className={getValidClassNames(styles.errorMessage, className)}>
      {hasError && error}
    </span>
  );
```

frontend/src/libs/components/error-message/styles.module.scss
```diff
@@ -0,0 +1,7 @@
+.errorMessage {
+  margin-top: 5px;
+  color: var(--red);
```
andanf-e, line 4:
```suggestion
  color: var(--red);
  min-height: <desired_height>
```

andanf-e (changes requested):

andanf-e (approved):

NazarKovtun1 (approved):

mhrechkin (approved):

PR #68, opened 2023-08-27
wr-65: add comment card component

OlenaMohylivska29:
Знімок екрана 2023-09-02 193331


frontend/src/libs/components/comment-card/comment-card.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { text, createdAt } = comment;
+  const userName = `${firstName} ${lastName}`;
+
+  return (
+    <section className={styles.comment}>
+      <div className={styles.header}>
+        <div className={styles.userInfo}>
+          <Link to={AppRoute.PROFILE}>
+            <Avatar username={userName} avatarUrl={null} />
+          </Link>
```
andanf-e: In comment it should not be a link - just avatar
```suggestion
            <Avatar username={userName} avatarUrl={null} />
```

andanf-e (changes requested):

frontend/src/libs/components/comments/comments.tsx
```diff
@@ -0,0 +1,30 @@
...
+  userId: 5,
+  articleId: 1,
+  text: `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Semper arcu natoque ultricies
+  lobortis sit sit vulputate. Commodo dictumst neque in ultricies eros sed iaculis eget. Netus
+  fames vulputate vitae nunc, orci ut. Vitae, neque, pharetra imperdiet nunc, commodo at.
+  Ultrices nibh sit a justo ac.`,
+  createdAt: '2 hours ago',
+};
+
+const Comments: React.FC = () => {
```
mhrechkin:
```suggestion
const Comments: React.FC = () => (
    <div className={styles.container}>
      <CommentCard user={MOCKED_USER} comment={MOCKED_COMMENT} />
    </div>
  );
```

mhrechkin (changes requested):

frontend/src/libs/types/comment.type.ts
drMangust, line 1: I see it already [here](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/100/files#diff-40b67043b46013cc7dc42989b046d69abda4f0a1ac5a95067ad375e2e049c692), look, maybe it would be better for you to take it from there

hnoievets: Thanks

hnoievets: This type doesn't have redundant props at the moment, and I think it would be better to change it after merging wr-73 branch

frontend/src/libs/components/comment-card/comment-card.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type Properties = {
+  user: UserAuthResponseDto;
+  comment: Comment;
+};
+
+const CommentCard: React.FC<Properties> = ({ user, comment }) => {
+  const { firstName, lastName } = user;
+  const { text, createdAt } = comment;
+  const userName = `${firstName} ${lastName}`;
```
andanf-e:
```suggestion
  const userName = getFullName(firstName, lastName);
```

frontend/src/libs/helpers/get-time-from-creation/get-time-from-creation.helper.ts
```diff
@@ -0,0 +1,30 @@
+import { ConvertedTimeUnit, TimeKeyword } from './enums/enums.js';
+
+const getTimeFromCreation = (isoDate: string): string => {
```
andanf-e: You've already added date-fns - just use `differenceIn{days/hours/minutes/seconds}` from it to return needed diff
It may looks like this
```
 const getDateDifferenceWithUnit = (date: string): { diff: number; unit: 'day' | 'hour' | 'minute' | 'second' } => {
    const currentDate = new Date();

    const diffInDays = differenceInDays(currentDate, date);
    if (diffInDays) {
      return { diff: diffInDays, unit: 'day' };
    }

    const diffInHours = differenceInHours(currentDate, date);
    if (diffInHours) {
      return { diff: diffInHours, unit: 'hour' };
    }

    const diffInMinutes = differenceInMinutes(currentDate, date);
    if (diffInMinutes) {
      return { diff: diffInMinutes, unit: 'minute' };
    }

    const diffInSeconds = differenceInSeconds(currentDate, date);
    return { diff: diffInSeconds, unit: 'second' };
 }
```

andanf-e (changes requested):

shared/src/libs/enums/time-unit.enum.ts
```diff
@@ -0,0 +1,8 @@
+const TimeUnit = {
+  DAY: 'day',
+  HOUR: 'hour',
+  MINUTE: 'min',
+  SECOND: 'sec',
```
andanf-e: I guess it would be better to have them in full form
```suggestion
  MINUTE: 'minute',
  SECOND: 'second',
```

frontend/src/libs/components/comment-card/comment-card.tsx
```diff
@@ -0,0 +1,50 @@
...
+  comment: CommentBaseResponseDto;
+};
+
+const CommentCard: React.FC<Properties> = ({ user, comment }) => {
+  const { firstName, lastName } = user;
+  const { text, createdAt } = comment;
+  const userName = getFullName(firstName, lastName);
+
+  const { diff, unit } = getDateDifferenceWithUnit(createdAt);
+  const publicationTime = `${diff} ${unit}`;
```
andanf-e: Let's also create local `getPublicationTime` helper as I guess you also need to provide ending `s` for diff `>1` (2 day`s` etc)

andanf-e (changes requested):

frontend/src/libs/components/comment-card/comment-card.tsx
hnoievets: Despite the design?
image

shared/src/libs/enums/time-unit.enum.ts
hnoievets: I didn't it because I didn`t see the full form anywhere in the design. Do I need to change it?

frontend/src/libs/components/comment-card/comment-card.tsx
andanf-e: Yep

shared/src/libs/enums/time-unit.enum.ts
andanf-e: Yes, I believe it would be more pretty

andanf-e (dismissed):

mhrechkin (approved):

andanf-e (approved):

PR #78, opened 2023-08-28
wr-12: get current user upon page refresh

canterbery:
1.  After refreshing the page, the user should still be logged in and able to access their account.
2.  After refreshing the page, the user data should be refetched from stored in localStorage token.
3.  The user should be redirected to the login page when encountering an expired token upon page refresh.
4.  The user should be redirected to the login page when encountering an invalid token upon page refresh.
5.  The user should be redirected to the login page when the token is unavailable upon page refresh.


shared/src/packages/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -2,6 +2,7 @@ const AuthApiPath = {
   ROOT: '/',
   SIGN_UP: '/sign-up',
   SIGN_IN: '/sign-in',
+  USER: '/user',
```
andanf-e:
```suggestion
  CURRENT: '/current',
```

frontend/src/libs/components/app/app.tsx
```diff
@@ -1,15 +1,48 @@
...
+  const hasUser = Boolean(user);
+  const hasToken = Boolean(storage.get(StorageKey.TOKEN));
+
+  if (hasUser && !hasToken) {
+    void dispatch(actions.logout());
+  }
+
+  const isLoading = !(
+    dataStatus === DataStatus.FULFILLED || dataStatus == DataStatus.REJECTED
+  );
```
andanf-e, line 25: Why not just?
```
  const isLoading = dataStatus === DataStatus.PENDING;
```

canterbery: If you do it this way it will go to sign-in page before request resolves and only then it will be redirected to root.   You can clearly see sign-in page blinking in the background on page refresh.   So I thought it's better to write check like that.

canterbery: As an option we can add  const [isLoading, setLoading] = useState(false); to component and handle request like this: 

  useEffect(() => {
    setLoading(true);
    if (hasToken) {
      dispatch(actions.getCurrentUser()).then(() => setLoading(false));
    }
  }, [hasToken, dispatch]);

andanf-e: It's because joined cases on reducer for dataStatuses per different actions,, actually they may be split up. But if so - then let's leave it as it is (with your implementation) and change the way how dataStatus is managed later on

frontend/src/libs/components/app/app.tsx
```diff
@@ -1,15 +1,48 @@
...
+    user: auth.user,
+    dataStatus: auth.dataStatus,
+  }));
+  const dispatch = useAppDispatch();
+  const hasUser = Boolean(user);
+  const hasToken = Boolean(storage.get(StorageKey.TOKEN));
+
+  if (hasUser && !hasToken) {
+    void dispatch(actions.logout());
+  }
```
andanf-e: Move it to `useEffect`

andanf-e (changes requested):

andanf-e (dismissed):

andanf-e (dismissed):

NazarKovtun1 (dismissed):

mhrechkin (dismissed):

andanf-e (approved):

PR #79, opened 2023-08-28
wr-66: Add password comparison upon Sign In

kovAnya:
Do I need to move InvalidCredentialsError to shared\src\libs\exceptions\invalid-credentials-error\invalid-credentials-error.exception.ts?


backend/src/packages/users/user.service.ts
```diff
@@ -51,6 +52,21 @@ class UserService implements IService {
...
+
+    const user = await this.userRepository.findByEmail(email);
+
+    return user
+      ? await this.encrypt.checkPassword(
+          password,
+          user.toNewObject().passwordHash,
+        )
+      : false;
+  }
```
andanf-e: And create new static getter `privateData` upon user entity that returns `passwordHash` and `passwordSalt`
```suggestion
  public async findPrivateData(id: number): Promise<UserPrivateData | null> {
    const user = await this.userRepository.find(id);

    if (!user) {
      return null;
    }

    return user.privateData;
  }
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -22,9 +22,14 @@ class AuthService {
 
     const user = await this.userService.findByEmail(email);
 
-    if (!user) {
-      throw new UserNotFoundError();
```
andanf-e, line 26: Turn back user exists assertion

backend/src/packages/auth/auth.service.ts
```diff
@@ -22,9 +22,14 @@ class AuthService {
...
+    const checkUserPassword = user
+      ? await this.userService.checkUserPassword(userSignInDto)
+      : null;
+
+    if (!user || !checkUserPassword) {
+      throw new InvalidCredentialsError();
```
andanf-e:
```suggestion
     const userPrivateData = (await this.userService.findPrivateData(
      user.id,
    )) as UserPrivateData;

    const hasSamePassword = await this.encrypt.compare({
      data: password,
      salt: userPrivateData.passwordSalt,
      passwordHash: userPrivateData.passwordHash,
    });

    if (!hasSamePassword) {
        throw new InvalidCredentialsError();
    }
```

backend/src/libs/packages/encrypt/encrypt.package.ts
```diff
@@ -10,6 +10,13 @@ class Encrypt implements IEncrypt {
   public async encrypt(password: string, salt: string): Promise<string> {
     return await hash(password, salt);
   }
+
+  public async checkPassword(
+    password: string,
+    hashPassword: string,
+  ): Promise<boolean> {
+    return await compare(password, hashPassword);
+  }
```
andanf-e:
```suggestion
  public async compare({
    passwordToCompare,
    salt,
    passwordHash,
  }: {
    passwordToCompare: string;
    salt: string;
    passwordHash: string;
  }): Promise<boolean> {
    const hashToCompare = await this.encrypt(data, salt);

    return hashToCompare === passwordHash;
  }
```

andanf-e (changes requested):

backend/src/packages/auth/auth.service.ts
kovAnya: @andanf-e So the user will know if the email is wrong before comparing the password? I assumed that the user just needs to know that the email and password match, not that the email doesn't exist. Or should I change the UserNotFoundError's message?

mhrechkin: If user not found, the error should be shown before password comparison.

mhrechkin (changes requested):

backend/src/packages/users/user.service.ts
```diff
@@ -51,6 +52,14 @@ class UserService implements IService {
     };
   }
 
+  public async findPrivateData(id: number): Promise<UserPrivateData | null> {
+    const user = await this.userRepository.find(id);
+    if (!user) {
+      return null;
+    }
```
andanf-e, line 60:
```suggestion

        if (!user) {
      return null;
    }
    
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -8,23 +12,40 @@ import {
...
+      user.id,
+    )) as UserPrivateData;
+    const hasSamePassword = await this.encrypt.compare({
+      passwordToCompare: password,
+      salt: userPrivateData.passwordSalt,
+      passwordHash: userPrivateData.passwordHash,
+    });
+    if (!hasSamePassword) {
+      throw new InvalidCredentialsError();
+    }
```
andanf-e:
```suggestion
    const userPrivateData = (await this.userService.findPrivateData(
      user.id,
    )) as UserPrivateData;
    
    const hasSamePassword = await this.encrypt.compare({
      passwordToCompare: password,
      salt: userPrivateData.passwordSalt,
      passwordHash: userPrivateData.passwordHash,
    });
    
    if (!hasSamePassword) {
      throw new InvalidCredentialsError();
    }
```

backend/src/libs/packages/encrypt/encrypt.package.ts
```diff
@@ -10,6 +10,19 @@ class Encrypt implements IEncrypt {
...
+    passwordToCompare,
+    salt,
+    passwordHash,
+  }: {
+    passwordToCompare: string;
+    salt: string;
+    passwordHash: string;
+  }): Promise<boolean> {
+    const hashToCompare = await this.encrypt(passwordToCompare, salt);
+    return hashToCompare === passwordHash;
```
andanf-e, line 25: Without any spacing dividers it's not the easiest thing to read
```suggestion

    return hashToCompare === passwordHash;
```

kovAnya: ok, got it

andanf-e (dismissed):

backend/src/packages/auth/auth.service.ts
```diff
@@ -8,23 +12,42 @@ import {
...
-    const { email } = userSignInDto;
+    const { email, password } = userSignInDto;
 
     const user = await this.userService.findByEmail(email);
-
     if (!user) {
       throw new UserNotFoundError();
     }
+
+    const userPrivateData = (await this.userService.findPrivateData(
```
PablloZz, line 37: Would it better to use destructuring here?

backend/src/packages/auth/auth.service.ts
```diff
@@ -8,23 +12,42 @@ import {
...
+      user.id,
+    )) as UserPrivateData;
```
PablloZz, line 39: Would it be better to use destructuring here?

kovAnya: I believe that it won't be useful a lot. This statement isn't too complex, is it?

andanf-e: Let's keep it as it is - like this we're grouping what values is internal, not from request payload

andanf-e (dismissed):

andanf-e (approved):

NazarKovtun1 (approved):

mhrechkin (approved):

PR #84, opened 2023-08-29
wr-75: Add articles page

### LINKED ISSUE #75: Add Articles page
image
**User story:**
As a logged user, I want to be able to explore and engage with articles, so that I can easily create new articles and stay informed about the latest content from other authors and share my own insights with the community.

**Acceptance criteria:**
- [x] Add page upon `/articles` route
- [x] Page should have 2 tabs: `Feed` and `My articles` (for now - may show both mocked Article cards
- [x] Page should be protected
After save article task would be merged - rely on data from BE + link it with articles fetch
- [x] Two tabs labeled “Feed” and “My Articles” should be visible on the Articles page.
- [x] The “Feed” tab should be the active tab by default.
- [x] User should be able to switch between tabs without any issues.
- [x] If an unauthorized user attempts to navigate to the Articles page he should be redirected to the login page.
- [x] The “Feed” tab should display all available articles, including the ones created by the user.
- [x] All available articles are displayed under the “Feed” tab in descending order (most recent first).
- [x] Articles created by the user are shown under the “My Articles” tab in descending order (most recent first).

PablloZz:


frontend/src/assets/css/scaffolding.scss
```diff
@@ -32,3 +32,7 @@ body {
   clip: rect(0 0 0 0);
   clip-path: inset(100%);
 }
+
+button {
+  padding: 0;
+}
```
andanf-e: Apply padding reset to all elements (as margin) - I've seen it in another PR, but just to don't miss it

frontend/src/index.tsx
```diff
@@ -12,10 +12,54 @@ import {
...
+  comments: '540',
+  views: '367K',
+  likes: '36K',
+  dislikes: '18K',
+  tags: MOCKED_TAGS,
+};
+
+const MOCKED_ARTICLES: ArticleType[] = Array.from({ length: 3 })
+  .fill(MOCKED_ARTICLE)
+  .map((article, index) => ({ ...(article as ArticleType), id: index }));
```
andanf-e: Move out mocked data from index file to articles component file - as we'd get real articles from store on that component

frontend/src/pages/articles/components/article-card/article-card.tsx
```diff
@@ -0,0 +1,82 @@
...
+          </div>
+          <div className={styles.reaction}>
+            <Icon iconName="like" className={styles.reactionIcon} />
+            <span className={styles.reactionCount}>{likes}</span>
+          </div>
+          <div className={styles.reaction}>
+            <Icon iconName="dislike" className={styles.reactionIcon} />
+            <span className={styles.reactionCount}>{dislikes}</span>
+          </div>
+          <Icon iconName="share" className={styles.reactionIcon} />
```
NazarKovtun1: Probably, we could render comments, views, likes and dislikes count components as a list because they all seem to have similar structure. @andanf-e @nickolay-grechkin what do you think?

andanf-e: yep 👍

frontend/src/pages/articles/articles.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Button
+          label="My articles"
+          onClick={handleShowArticles}
+          className={getValidClassNames(
+            styles.button,
+            currentPage === ActivePage.ARTICLES && styles.activeButton,
+          )}
+        />
+      </div>
+      <div className={styles.articles}>{renderArticles}</div>
```
andanf-e: Let's use nested routes to switch between tabs

andanf-e (changes requested):

frontend/src/pages/articles/articles.tsx
PablloZz: If we get the articles from the state, we don't need to specify any components under the paths of the nested routes to use them with outlet.
So can I ignore this warning: ?
image

andanf-e: It's hard to say if it's okay - let's take a look on how it was done, and it would be more clear

frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -3,7 +3,10 @@ const AppRoute = {
   SIGN_IN: '/sign-in',
   SIGN_UP: '/sign-up',
   PROFILE: '/profile',
+  ARTICLES: '/articles',
   ARTICLE: '/articles/:id',
+  FEED: 'feed',
+  MY_ARTICLES: 'my-articles',
```
andanf-e: Move it to `ArticleSubRoute` enum

frontend/src/pages/articles/components/article-card/article-card.tsx
```diff
@@ -1,71 +1,31 @@
...
-  const tags = propertyTags.map((tag) => <Tag key={tag.id} name={tag.name} />);
-
+  const renderTags = tags.map((tag) => <Tag key={tag.id} name={tag.name} />);
+  const renderReactions = REACTIONS_LIST.map((reaction) => (
+    <Reaction
+      key={reaction.reaction}
+      iconName={reaction.iconName}
+      reactions={article[reaction.reaction]}
+    />
+  ));
```
andanf-e: I believe Nazar meant to use ul and list items per different staff
Also it's not very clear how it looks for now - make them render separately by in `<ul><li>...</ul>` form
Same for `tags`

frontend/src/pages/articles/articles.tsx
```diff
@@ -0,0 +1,99 @@
...
+            label="My articles"
+            onClick={handleShowMyArticles}
+            className={getValidClassNames(
+              styles.button,
+              activePage === AppRoute.MY_ARTICLES && styles.activeButton,
+            )}
+          />
+        </Link>
+      </div>
+      <div className={styles.articles}>{renderArticles}</div>
```
andanf-e: Call this file and component `ArticlesPage`, create separate components for `ArticlesFeed` and `MyArticles` (for now they would be the same,, export them from `articles` file (export `ArticlesFeed`, `MyArticles` and `ArticlesPage`)
```suggestion
      <div className={styles.articles}><Outlet /></div>
```

frontend/src/pages/articles/articles.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <Link to={AppRoute.MY_ARTICLES}>
+          <Button
+            label="My articles"
+            onClick={handleShowMyArticles}
+            className={getValidClassNames(
+              styles.button,
+              activePage === AppRoute.MY_ARTICLES && styles.activeButton,
+            )}
+          />
+        </Link>
```
andanf-e: Should be just links, not buttons

frontend/src/pages/articles/articles.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  useEffect(() => {
+    if (activePage === AppRoute.FEED) {
+      return handleShowFeed();
+    }
+
+    if (activePage === AppRoute.MY_ARTICLES) {
+      return handleShowMyArticles();
+    }
+  }, [handleShowFeed, handleShowMyArticles, activePage]);
```
andanf-e: I believe with nested routing it's not needed

frontend/src/index.tsx
```diff
@@ -54,6 +56,22 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
...
+                  <ProtectedRoute>
+                    <Articles />
+                  </ProtectedRoute>
+                ),
+                children: [
+                  {
+                    path: AppRoute.FEED,
+                  },
+                  {
+                    path: AppRoute.MY_ARTICLES,
```
andanf-e: And corresponding ArticlesFeed and MyArticles components as elements

andanf-e (changes requested):

mhrechkin (dismissed):

frontend/src/pages/articles/components/article-card/article-card.tsx
PablloZz: Can I create common tags component instead of tag one?
Or would it be better to remove the tag component when the article view task is merged, since that component is used there

andanf-e: Just use `<ul><li>...</ul>` as later on we'd make some of them interactive (likes/dislikes/comments) and they'd have separate logic -- there is no need to create something common

frontend/src/pages/articles/components/reactions/reactions.tsx
```diff
@@ -0,0 +1,38 @@
...
+    </li>
+    <li className={styles.reaction}>
+      <Icon iconName="like" className={styles.reactionIcon} />
+      <span className={styles.reactionCount}>{likes}</span>
+    </li>
+    <li className={styles.reaction}>
+      <Icon iconName="dislike" className={styles.reactionIcon} />
+      <span className={styles.reactionCount}>{dislikes}</span>
+    </li>
+  </ul>
```
andanf-e: Should be in article-card component

andanf-e (changes requested):

frontend/src/pages/articles/components/reactions/reactions.tsx
PablloZz: What about the tags, do I need to move them here too? I'm a bit confused because in the previous task I was told that it's better to move the tags to a separate component.

andanf-e: Tags may be in separate component - as they're just displaying some sort of information (and will do it later on)

frontend/src/pages/articles/articles-page.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const activePage = pathname.split('/').pop();
+
+  return (
+    <div className={styles.wrapper}>
+      <div className={styles.tabsWrapper}>
+        <Link
+          to={ArticleSubRoute.FEED}
+          className={getValidClassNames(
+            styles.tab,
+            activePage === ArticleSubRoute.FEED && styles.activeTab,
```
andanf-e: I believe `Navlink` as an option injects a callback passing for `className` prop with `isActive` argument you may check if we can accomplish active link styling with this approach
Like
```
pass optional 'activeClassName' style for Link component 
And in Link component change className passing to something like 
({ isActive }) => getValidClassNames(styles.link, className, isActive && activeClassName)
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,21 +1,36 @@
...
-);
+const Link: React.FC<Properties> = ({
+  children,
+  className,
+  activeClassName,
+  to,
+}) => {
+  const handleGetClassesByLinkState = useCallback(
+    ({ isActive }: { isActive: boolean }): string =>
+      getValidClassNames(styles.link, className, isActive && activeClassName),
```
andanf-e: Without brackets prettier may shuffle line as it wants (and create unnecessary diff without any reason)
```suggestion
  const handleGetClassesByLinkState = useCallback(
    ({ isActive }: { isActive: boolean }): string => {
      return getValidClassNames(styles.link, className, isActive && activeClassName);
    },
```

andanf-e (approved):

mhrechkin (approved):

PR #85, opened 2023-08-29
wr-76: article prompt generation using openai

artemmatiushenko1:
This PR adds an API endpoint to get a random article prompt generated by openAI. I was playing with completion settings, and changing prompt, but the generated prompt elements are not very unique every time. Chat GPT prefers to generate a lot of science fiction and sci-fi genres 😅 Maybe you can take a look and suggest how to improve a prompt for open AI to get more diverse results


backend/src/libs/packages/openai/openai.package.ts
```diff
@@ -0,0 +1,48 @@
...
+      messages: [
+        {
+          'role': 'user',
+          'content': prompt,
+        },
+      ],
+    });
+
+    const [modelChoice] = modelResponse.choices;
+    return modelChoice?.message.content;
```
andanf-e: For narrowing returning result to `string | null`
```suggestion
    return modelChoice?.message.content ?? null;
```

NazarKovtun1 (dismissed):

backend/src/libs/packages/openai/openai.package.ts
```diff
@@ -0,0 +1,48 @@
+import { OpenAI as OpenAIClient } from 'openai';
+
+import { type IConfig } from '../config/config.js';
+import { type CompletionConfig } from './libs/types/completion-config.type.js';
+
+class OpenAIService {
+  private config: IConfig;
+  private openAIClient: OpenAIClient;
+
+  private DEFAULT_COMPLETION_MODEL = 'gpt-3.5-turbo';
```
andanf-e, line 10: Let's move model to env variables - we may swap between them to check the resulting prompts

artemmatiushenko1: Should I specify a default env variable value (for example `gpt-3.5-turbo`) or leave it as `null`?

andanf-e: Yep, let's specify default but let's specify instead of `gpt-3.5-turbo` some lower model like `babbage` or `ada`

artemmatiushenko1: @andanf-e I just checked the docs and it says that such models like Babbage and Ada are legacy now and have a little bit different API to make a prompt comparing to GPT-3.5+ models. So, should I create some flag env variable like USE_LEGACY_MODEL, and use the corresponding API based on the value of that flag then?
image

andanf-e: Sorry, if so, then lets keep it as it is 🥲

andanf-e (approved):

mhrechkin (approved):

PR #87, opened 2023-08-29
wr-72: Create genres table

IrynaK-a:


backend/src/db/migrations/20230828103237_create_genres_table.ts
```diff
@@ -0,0 +1,37 @@
+import { type Knex } from 'knex';
+
+import { GENRES } from './libs/constants/genres.js';
```
andanf-e: Place them directly in migration file

backend/src/db/migrations/libs/constants/genres.ts
```diff
@@ -0,0 +1,24 @@
+const GENRES = [
+  { name: 'Science Fiction', key: 'sci-fi' },
```
andanf-e: Use snake_case for keys, change it for all keys with kebab-case
```suggestion
  { name: 'Science Fiction', key: 'sci_fi' },
```

andanf-e (changes requested):

andanf-e (dismissed):

backend/src/db/migrations/20230828103237_create_genres_table.ts
```diff
@@ -0,0 +1,58 @@
...
+  { name: 'Non-fiction', key: 'non_fiction' },
+  { name: 'Paranormal', key: 'paranormal' },
+  { name: 'Satire', key: 'satire' },
+  { name: 'Young Adult', key: 'ya_fiction' },
+  { name: 'Crime', key: 'crime' },
+  { name: 'Drama', key: 'drama' },
+  { name: 'Science', key: 'science' },
+  { name: 'Urban Fantasy', key: 'urban_fantasy' },
+  { name: 'Memoir', key: 'memoir' },
+];
```
NazarKovtun1, line 24: Should not it be moved to seeds?

backend/src/packages/genres/genre.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private constructor({
+    id,
+    name,
+    key,
+  }: {
+    id: number;
+    name: string;
+    key: string;
+  }) {
```
NazarKovtun1: Please, create a type for it and put it right in this file (right before GenreEntity)

NazarKovtun1: Also, check all other places, where this case with type is not covered (not just method parameters but return types as well)

NazarKovtun1 (changes requested):

backend/src/db/migrations/20230828103237_create_genres_table.ts
andanf-e: Nope, seeds are not the best choice for deployed app data, as it should be consistent

backend/src/packages/genres/genre.entity.ts
IrynaK-a: Do you mean in other entities too?

andanf-e (dismissed):

backend/src/packages/genres/genre.entity.ts
mhrechkin: @WilhelmWesser i think we use the same approach in all entities, so there is no need to change it in this pr

NazarKovtun1: Okey then

NazarKovtun1 (dismissed):

PR #89, opened 2023-08-29
wr-88: Public/Protected routes changes 

IrynaK-a:


frontend/src/pages/auth/auth.tsx
andanf-e, line 1: These changes is not needed
Check Public Route component in this PR
https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/78/files#diff-081b8e6bd64aab9a561d807c75a72d502737edbc025e4c4cd96102eab0e359bf

andanf-e (changes requested):

frontend/src/pages/auth/auth.tsx
canterbery: I made it protected because it was the only page avaliable to test access on.  But it should be made public when we add root page for authorized users to the project.  So if user profile is that root page we should make landing page public and change default redirect path in public route component to user profile

andanf-e: @canterbery
Comment in `auth` component was only about the way how to redirect authorized user from public route (to use your already implemented approach)
I understand, and that's why - after profile + other tasks merging, I've created this ticket

canterbery: Sorry , hasn't checked ticket itself before commenting.

Issue #90, opened 2023-08-29
Create inner prompt for OpenAI call to specify created article genre


andanf-e:
When article is submitted:
- [x] Take article text and pass it to OpenAI prompt to get genre of that story (may return an array of genres)
- [x] As of initial implementation just save one of returned genres (if present) as article's genre
- [x] If our app's `genres` does not contain genre that is applied for article, then save it in `genres`

Neyrinka:
**Actual Result:** The genre is not displayed when the AI cannot determine the genre of the article

**Expected Result:** The genre should be set as "unknown" when the AI cannot determine the genre of the article.

PR #96, opened 2023-08-29
wr-90: detect genre from article text

artemmatiushenko1:
This PR adds implementation for article's genre detection using OpenAI


drMangust:
> @[artemmatiushenko1](https://github.com/artemmatiushenko1) artemmatiushenko1 linked an issue [29 minutes ago](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/96#event-10225937113) that may be closed by this pull request
> https://github.com/BinaryStudioAcademy/bsa-2023-writorium/issues/90

It's better not to do that. I did and it automatically closes the issue, and we don't need that. I thought it was convenient to have access to both the issue and the PR from both places, but this automatically moves the issue to done. Maybe this can be customized.

artemmatiushenko1:
@drMangust Ok! Thanks for the notice 🙂

backend/src/db/migrations/20230822132525_create_articles_table.ts
```diff
@@ -28,7 +28,7 @@ const up = (knex: Knex): Promise<void> => {
       .references(ColumnName.ID)
       .inTable(TableName.USERS);
     table.integer(ColumnName.PROMPT_ID).unsigned().nullable();
-    table.integer(ColumnName.GENRE_ID).unsigned();
+    table.integer(ColumnName.GENRE_ID).unsigned().nullable();
```
andanf-e: Create new migration to alter table with nullable column
It won't be applied on deployed env as this migration has been already been run no deployed app database
It's almost always a bad idea to edit old migrations 🥲 
```suggestion
    table.integer(ColumnName.GENRE_ID).unsigned();
```

backend/src/packages/articles/article.service.ts
```diff
@@ -11,9 +19,63 @@ import {
...
+  private async getGenreIdForArticle(
+    articleText: string,
+  ): Promise<number | null> {
+    const detectedGenre = await this.detectArticleGenreFromText(articleText);
+
+    if (!detectedGenre) {
+      return null;
+    }
+
+    const detectedGenreKey = convertPlainStringToSnakeCase(detectedGenre);
```
andanf-e: Can we just ask in prompt to return snake_cased key together with readable name?

andanf-e (changes requested):

backend/src/db/migrations/20230822132525_create_articles_table.ts
artemmatiushenko1: Got it, thanks!

IrynaK-a: maybe it isn't necessary to create a new migration because it is nullable by default and does not change anything?

artemmatiushenko1: @IrynaK-a I guess it's just for code consistency

andanf-e: Yeah, I've missed thing with default nullable 😅
Just recognized changed migration - let's not create new migration (but remove changes from existing one)

backend/src/packages/articles/article.service.ts
```diff
@@ -1,19 +1,74 @@
...
+      prompt: getDetectArticleGenrePrompt(text),
+    });
+
+    if (!generesJSON) {
+      return null;
+    }
+
+    const parsedGenres = safeJSONParse<DetectedArticleGenre[]>(generesJSON);
+
+    return parsedGenres?.[0] ?? null;
```
andanf-e: To avoid magic number usage + FYI about typo
```suggestion
    const [firstParsedGenre] = safeJSONParse<DetectedArticleGenre[]>(genresJSON);

    return firstParsedGenre ?? null;
```

backend/src/packages/articles/article.service.ts
```diff
@@ -1,19 +1,75 @@
...
+
+    if (existingGenre) {
+      return existingGenre.toObject().id;
+    } else {
+      const newGenreEntity = await this.genreRepository.create(
+        GenreEntity.initializeNew(detectedGenre),
+      );
+
+      return newGenreEntity.toObject().id;
+    }
```
andanf-e: Else block is not necessary as you have return in if block
```suggestion
    if (existingGenre) {
      return existingGenre.toObject().id;
    }
    
    const newGenreEntity = await this.genreRepository.create(
        GenreEntity.initializeNew(detectedGenre),
      );

      return newGenreEntity.toObject().id;
```

shared/src/packages/articles/libs/validation-schemas/article-create.validation-schema.ts
```diff
@@ -28,7 +28,7 @@ const articleCreate = joi.object<ArticleRequestDto, true>({
       'string.empty': ArticleValidationMessage.ARTICLE_TEXT_REQUIRE,
     }),
   promptId: joi.number().integer().positive().required().allow(null),
-  genreId: joi.number().integer().positive().required().messages({
+  genreId: joi.number().integer().positive().messages({
```
andanf-e: I believe `article genre id required` message on empty string is not necessary now + add `.allow(null)`

andanf-e (dismissed):

backend/src/packages/genres/genre.entity.ts
```diff
@@ -1,42 +1,54 @@
...
       name,
       key,
     });
   }
 
-  public toObject(): Genre {
+  public static initializeNew({
+    name,
+    key,
+  }: Omit<GenreEntityType, 'id'>): GenreEntity {
```
NazarKovtun1, line 34: Please, create this type separately
type GenreEntityPayloadType = Omit<GenreEntityType, 'id'>
move it before the class and use here

backend/src/packages/articles/article.service.ts
```diff
@@ -1,19 +1,75 @@
...
     this.articleRepository = articleRepository;
+    this.openAIService = openAIService;
+    this.genreRepository = genreRepository;
+  }
+
+  public async detectArticleGenreFromText(
+    text: string,
+  ): Promise<DetectedArticleGenre | null> {
+    const genresJSON = await this.openAIService.createCompletion({
+      temperature: 0,
```
NazarKovtun1: Magic number? @andanf-e @nickolay-grechkin

backend/src/packages/genres/genre.entity.ts
andanf-e: Not sure if it's needed for such case

backend/src/packages/articles/article.service.ts
andanf-e: **B31** yep, missed

artemmatiushenko1: @andanf-e will it be ok if I move the whole completion config to `getDetectArticleGenrePrompt` helper and name it `getDetectArticleGenreCompletionConfig` instead ?

andanf-e: Yes, but still move temperature value to constant in that helper

andanf-e (approved):

mhrechkin (approved):

PR #97, opened 2023-08-29
wr-92: Fix background images paths on auth pages

IrynaK-a:


frontend/vite.config.ts
```diff
@@ -24,6 +25,11 @@ const config = ({ mode }: ConfigEnv): ReturnType<typeof defineConfig> => {
         },
       },
     },
+    resolve: {
+      alias: {
+        '@img': path.resolve(__dirname, './src/assets/img'),
```
andanf-e: Maybe it would be better to link assets folder, in case if we'd need to get path for anything else
```suggestion
        '@assets': path.resolve(__dirname, './src/assets'),
```

frontend/vite.config.ts
```diff
@@ -24,6 +25,12 @@ const config = ({ mode }: ConfigEnv): ReturnType<typeof defineConfig> => {
...
+        '@assets': path.resolve(__dirname, './src/assets'),
```
andanf-e, line 30: I mean use it with `@assets` instead of `@img`

andanf-e (changes requested):

frontend/vite.config.ts
IrynaK-a: Oh, sorry, got it

andanf-e (approved):

NazarKovtun1 (approved):

PR #98, opened 2023-08-30
wr-82: Attach authorized user to the header

### LINKED ISSUE #82: Attach authorized user to header
**User story:**
As a logged-in user of the application, I want to see my notifications and have my user avatar displayed in the Header, So that I can easily access and stay informed about any new notifications and personalize my experience.

**Acceptance criteria:**
- [ ] Authorized user from store should be passed to Header component. 
- [ ] When a user log in, the Header should display his notifications count and his user avatar.
- [ ] If a user receive new notifications, the Header should update the notification count in real-time.
- [ ] If a user change his avatar, the new avatar should be immediately reflected in the Header. **This will be fixed in** https://github.com/BinaryStudioAcademy/bsa-2023-writorium/issues/172
- [ ] When a user log out, the Header should no longer display his notifications and user avatar.

PablloZz:


frontend/src/libs/components/app/app.tsx
```diff
@@ -37,7 +37,7 @@ const App: React.FC = () => {
   return (
     <Loader isLoading={isLoading}>
       <>
-        <Header user={null} />
+        {hasUser && <Header user={user} />}
```
andanf-e: I believe it's not necessary if we have separate check that user exists inside header component
So lets deal with a single approach - we are checking if user exists here OR inside header component (not in both places)

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,10 +1,11 @@
 import { Avatar, Link } from '~/libs/components/components.js';
 import { AppRoute } from '~/libs/enums/enums.js';
+import { type UserAuthResponseDto } from '~/packages/users/users.js';
 
 import styles from './styles.module.scss';
 
 type Properties = {
-  user: null;
+  user: UserAuthResponseDto | null;
```
andanf-e, line 9: We're now having real user from store - pass his own data to header instead of placeholder `username`

andanf-e (changes requested):

andanf-e (approved):

mhrechkin (approved):

PR #115, opened 2023-08-31
wr-106: Article page doesn't match the design

AnastasiaR2:
Created container and styles for article component to place the author section on the right side form the article.

image

### DIFF
--- a/frontend/src/libs/components/components.ts
+++ b/frontend/src/libs/components/components.ts
@@ -6,6 +6,7 @@ export { ErrorMessage } from './error-message/error-message.js';
 export { Header } from './header/header.js';
 export { Icon } from './icon/icon.js';
 export { Input } from './input/input.js';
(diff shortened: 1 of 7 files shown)


andanf-e (changes requested):
image
From the screenshot provided it looks like articles page still does not have inline padding as on the design + I believe that inline padding should be the same for all pages (except Sign In\Sign Up with it's own layout)
Also top padding may also be added for all pages

AnastasiaR2:
It is required to create paddings separately on every page, or some general styles for all pages?

andanf-e:
> It is required to create paddings separately on every page, or some general styles for all pages?

General, mb create some Layout component, or smth like that and wrap pages that need to have it

andanf-e (approved):
But update screenshot please

NazarKovtun1 (approved):

PR #119, opened 2023-08-31
wr-107: Add Write link to header

hnoievets:
- [x]  Next to user profile avatar add Write link
- [x]  Write link should point to page with Article creation form

Didn't create an articleCreate page because Anastasia is doing it.
Removed `color` from Link styles because there were difficulties with changing the font color for the link, because of this, the color of the logo was wrong.

image


frontend/src/libs/components/header/header.tsx
```diff
@@ -16,12 +16,20 @@ const Header: React.FC<Properties> = ({ user }) => (
...
-          <Avatar
-            username={getFullName(user.firstName, user.lastName)}
-            avatarUrl={null}
-          />
-        </Link>
+        <div className={styles.rightSide}>
+          <Link to={AppRoute.CREATE_ARTICLE}>
+            <Button
+              label="Write"
+              className={getValidClassNames(styles.linkBtn, styles.writeLink)}
```
andanf-e: Just add custom styling for Link to style it as button

NazarKovtun1 (dismissed):

frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -5,6 +5,7 @@ const AppRoute = {
   PROFILE: '/profile',
   ARTICLES: '/articles',
   ARTICLE: '/articles/:id',
+  CREATE_ARTICLE: '/create',
```
andanf-e:
```suggestion
  CREATE_ARTICLE: '/articles/create',
```

andanf-e (changes requested):

frontend/src/libs/components/link/styles.module.scss
```diff
@@ -1,4 +1,3 @@
 .link {
-  color: inherit;
```
andanf-e, line 2: Why do we need to remove it?

hnoievets: Removed `color` from Link styles because there were difficulties with changing the font color for the link, because of this, the color of the logo was wrong.

hnoievets: If I'm not mistaken, we need to change the color of the parent component instead of changing it directly

andanf-e: I believe `color: inherit` takes a color for link from the upper parent styles

hnoievets: When using `color: inherit` the styles will look like this(also changed for logo because now it's black):
image
image

And Link prop `className` can't change the color of font, if it's ok, we don't need to remove it

andanf-e: Okay, you mean that with `color: inherit` font color applies only from parent element color styling - if so, then let's remove it and pass color directly through class for link element 👍

andanf-e (dismissed):

frontend/src/libs/components/link/styles.module.scss
hnoievets: Yes, that's what I mean

andanf-e (approved):

mhrechkin (approved):

PR #125, opened 2023-09-01
wr-111: Add Prompt Generation Component to Create Article page

AnastasiaR2:
1. Created prompts table, model, entity, repository.
2. Implemented logic to save prompt on BE.
3. Added Create Article Page and routes for it, updated Article folder.
4. If a user generates a prompt - the promptId and genreId from it are added upon the article submission.

image


frontend/src/pages/article/components/article-form/article-form.tsx
```diff
@@ -1,26 +1,48 @@
...
+  useEffect(() => {
+    if (generatedPrompt) {
+      void dispatch(
+        promptsActions.createPrompt({
+          type: PromptType.MANUAL,
+          ...generatedPrompt,
+        }),
+      );
+    }
+  }, [dispatch, generatedPrompt]);
```
AnastasiaR2, line 44: useEffect is used here to save the prompt to the database if there is a generated prompt. But I'm not sure if it's correct. I tried to do that in handleArticleSubmit, but createdPrompt will be null after the dispatch createPrompt, I also tried to use async/await, but it doesn't help. Please suggest something here

AnastasiaR2: useEffect is used here to save the prompt to the database if there is a generated prompt. But I'm not sure if it's correct. I tried to do that in handleArticleSubmit, but createdPrompt will be null after the dispatch createPrompt, I also tried to use async/await, but it doesn't help. Please suggest something here

artemmatiushenko1: Maybe just move prompt creation logic to `articlesActions.createArticle` and pass `generatedPrompt` in a payload as well?

backend/src/db/migrations/20230901145235_create_prompts_table.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const up = (knex: Knex): Promise<void> => {
+  return knex.schema.createTable(TableName.PROMPTS, (table) => {
+    table.increments(ColumnName.ID).primary();
+    table.text(ColumnName.CHARACTER).nullable();
+    table.text(ColumnName.SETTING).nullable();
+    table.text(ColumnName.SITUATION).nullable();
+    table.text(ColumnName.PROP).nullable();
+    table
+      .enu(ColumnName.TYPE, [PromptType.DAILY, PromptType.MANUAL])
```
andanf-e: That was valid for knex, but just to make it look more readabale
```suggestion
      .enum(ColumnName.TYPE, [PromptType.DAILY, PromptType.MANUAL])
```

shared/src/libs/helpers/convert-plain-string-to-snake-case/convert-plain-string-to-snake-case.helper.ts
```diff
@@ -0,0 +1,8 @@
+const convertPlainStringToSnakeCase = (input: string): string => {
+  return input
+    .split(' ')
+    .map((word) => word.replaceAll(/-+/g, '_').toLowerCase())
+    .join('_');
+};
```
andanf-e: Maybe let's use lodash.snakecase to don't give a care about any cases with wording to snake_case

backend/src/packages/prompts/prompt.controller.ts
```diff
@@ -38,6 +68,16 @@ class PromptsController extends Controller {
       method: 'GET',
       handler: () => this.generate(),
     });
+    this.addRoute({
+      path: PromptsApiPath.ROOT,
+      method: 'POST',
+      handler: (options) =>
```
andanf-e, line 78: Add validation for it

andanf-e (changes requested):

frontend/src/slices/prompts/prompts.slice.ts
```diff
@@ -1,18 +1,23 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createSlice, isAnyOf } from '@reduxjs/toolkit';
 
 import { DataStatus } from '~/libs/enums/enums.js';
 import { type ValueOf } from '~/libs/types/types.js';
-import { type GenerateArticlePromptResponseDto as Prompt } from '~/packages/prompts/prompts.js';
+import {
+  type GenerateArticlePromptResponseDto as generatedPrompt,
+  type PromptBaseResponseDto as createdPrompt,
+} from '~/packages/prompts/prompts.js';
```
PablloZz: Maybe it will be better to rename types to UpperCase to be consistent with types naming

frontend/src/pages/article/components/article-form/article-form.tsx
AnastasiaR2: Thanks, will do)

backend/src/packages/articles/article.service.ts
```diff
@@ -97,8 +97,8 @@ class ArticleService implements IService {
   public async create(
     payload: ArticleCreateDto,
   ): Promise<ArticleBaseResponseDto> {
-    const genreId = await this.getGenreIdForArticle(payload.text);
-
+    const genreId =
+      payload.genreId ?? (await this.getGenreIdForArticle(payload.text));
```
andanf-e: Let's move it to helper - to make it more readable

backend/src/packages/prompts/prompt.service.ts
```diff
@@ -1,19 +1,35 @@
+import snakeCase from 'lodash.snakecase';
```
andanf-e: Move it to helper function, that will wrap lodash.snakeCase call

frontend/src/pages/article/components/article-form/article-form.tsx
```diff
@@ -32,9 +40,24 @@ const ArticleForm: React.FC = () => {
...
+                  ...generatedPrompt,
+                },
+              }),
+            )
+          : void dispatch(
+              articlesActions.createArticle({
+                articlePayload: updatedPayload,
+                generatedPrompt: null,
+              }),
+            );
```
andanf-e: Or even move getGeneratedPrompt to helper fn to make it more readable
```suggestion
        void dispatch(
              articlesActions.createArticle({
                articlePayload: updatedPayload,
                generatedPrompt: generatedPrompt
                  ? {
                    type: PromptType.MANUAL,
                    ...generatedPrompt,
                  }
                  : null,
              }),
            );
```

frontend/src/slices/articles/actions.ts
```diff
@@ -31,12 +32,25 @@ const fetchOwn = createAsyncThunk<
...
+  async ({ articlePayload, generatedPrompt }, { extra }) => {
+    const { articleApi, promptApi } = extra;
 
-  return articleApi.create(articlePayload);
-});
+    if (generatedPrompt) {
+      const { id: promptId, genreId } = await promptApi.create(generatedPrompt);
+
+      articlePayload.genreId = genreId;
+      articlePayload.promptId = promptId;
```
andanf-e: Let's better make separated call, instead of mutating payload
```suggestion
      return await articleApi.create({
        ...articlePayload,
        genreId,
        promptId
      });
```

package.json
```diff
@@ -30,6 +30,7 @@
     "@commitlint/cli": "17.7.1",
     "@commitlint/config-conventional": "17.7.0",
     "@ls-lint/ls-lint": "2.0.1",
+    "@types/lodash.snakecase": "4.1.7",
```
andanf-e: Can't find where dependency is installed, and I guess it should be for backend workspace

andanf-e (changes requested):

backend/src/packages/articles/article.service.ts
AnastasiaR2: Do you mean creating something like this?
```
const getGenreId = async (payload: ArticleRequestDto): Promise<number | null> =>
  payload.genreId ?? (await this.getGenreIdForArticle(payload.text));
  
const genreId = await getGenreId(payload);
```

package.json
AnastasiaR2: lodash.snakecase is installed in common node modules, should I move it to the backend workspace, as well as @types/lodash.snakecase?

andanf-e: Diff doesn't show package installed anywhere + yep, it should be part of backend deps, as it's used on backend

backend/src/packages/articles/article.service.ts
andanf-e: Something like this
```
const getGenreIdToSet = async ({ genreId, text }: ArticleRequestDto): Promise<number | null> => {
  if (genreId) {
    return genreId;
  }
  
  return this.getGenreIdForArticle(text);
};
```

backend/package.json
```diff
@@ -47,9 +47,11 @@
   "devDependencies": {
     "@types/bcrypt": "5.0.0",
     "@types/convict": "6.1.3",
+    "@types/lodash.snakecase": "4.1.7",
     "@types/nodemailer": "6.4.9",
     "@types/pug": "2.0.6",
     "@types/swagger-jsdoc": "6.0.1",
+    "lodash.snakecase": "4.1.1",
```
andanf-e: It's not a dev dependency

backend/src/libs/helpers/convert-to-snake-case/convert-to-snake-case.helper.ts
```diff
@@ -0,0 +1,5 @@
+import snakeCase from 'lodash.snakecase';
+
+const convertToSnakeCase = (input: string): string => snakeCase(input);
```
andanf-e:
```suggestion
const convertToSnakeCase = (input: string): string => {
  return snakeCase(input);
};
```

frontend/src/assets/css/variables.scss
```diff
@@ -38,3 +37,9 @@ input {
   --input-text: rgb(40 40 40 / 50%);
   --input-placeholder-color: #9da4ae;
 }
+
+textarea {
+  --textarea-bg: var(--pale-gray);
+  --textarea-text: var(--dark-blue);
+  --textarea-shadow-color: rgba(53 53 53 / 20%);
+}
```
andanf-e: I guess it can be removed as there is already changes with RTE
```suggestion

```

andanf-e (changes requested):

andanf-e (approved):

NazarKovtun1 (approved):

PR #126, opened 2023-09-02
wr-108: Create dropdown with actions on header user avatar

kovAnya:


frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,95 @@
...
+  return (
+    <>
+      {user && (
+        <div className={styles.outerheader}>
+          <header className={styles.header}>
+            <Link to={AppRoute.ROOT} className={styles.logo}>
+              WRITORIUM
+            </Link>
+
+            <div className={styles.avatarWrapper} ref={avatarReference}>
```
andanf-e: Why not just wrapping it with button?

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,95 @@
...
+              onClose={closeDropdown}
+              className={`${styles.dropdownModal} ${
+                styles.dropdownModalStyle
+              } ${isOpen ? styles.open : ''}`}
+            >
+              <DropDown />
+            </Modal>
+          )}
+        </div>
+      )}
```
andanf-e: Create common Popover component that will take `content` prop (with staff that would be shown on dropdown) and `trigger` prop (with staff that would trigger Popover opening)

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,95 @@
...
+  const location = useLocation();
+
+  const [previousLocation, setPreviousLocation] = useState(location);
+
+  useEffect(() => {
+    if (isOpen && location !== previousLocation) {
+      handleToggleModalOpen();
+      setPreviousLocation(location);
+    }
+  }, [location, previousLocation, isOpen, handleToggleModalOpen]);
```
andanf-e: Not sure why do we need this

andanf-e (changes requested):

frontend/src/libs/components/header/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  trigger: {
+    handleToggleModalOpen: () => void;
+    isOpen: boolean;
+  };
+};
+
+const DropDown: React.FC<Properties> = ({ trigger }) => {
+  const dispatch = useAppDispatch();
+
+  const handleSignOut = useCallback((): void => {
```
artemmatiushenko1: I guess it would be better to name it as `handleLogout`, because you have a redux action and css class name using logout word, just to make it consistent

frontend/.eslintrc.yml
```diff
@@ -19,9 +19,7 @@ extends:
 rules:
   react/prop-types:
     - off
-  react/jsx-no-bind:
-    - error
-    - ignoreRefs: true
+  react/jsx-no-bind: ['error', { ignoreRefs: true, allowArrowFunctions: true }]
```
artemmatiushenko1: This is a `yml` file, so if I’m not mistaken it should be something like

```yml

react/jsx-no-bind:
  - error
  - ignoreRefs: true
    allowArrowFunctions: true
```

frontend/src/libs/components/header/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './styles.module.scss';
+
+type Properties = {
+  trigger: {
+    handleToggleModalOpen: () => void;
+    isOpen: boolean;
+  };
+};
+
+const DropDown: React.FC<Properties> = ({ trigger }) => {
```
artemmatiushenko1: I guess `Dropdown` would be better, because such name is commonly used for that component 🙂

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,56 @@
...
-          WRITORIUM
-        </Link>
-        <Link to={AppRoute.PROFILE}>
-          <Avatar
-            username={getFullName(user.firstName, user.lastName)}
-            avatarUrl={null}
+const Header: React.FC<Properties> = ({ user }) => {
+  const { handleToggleModalOpen, isOpen } = useModal();
+
+  const handleClickOnAvatar = useCallback((): void => {
```
artemmatiushenko1: `handleAvatarClick`

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,56 @@
...
+    if (!isOpen) {
+      handleToggleModalOpen();
+    }
+  }, [handleToggleModalOpen, isOpen]);
+
+  return (
+    <>
+      {user && (
+        <div className={styles.outerheader}>
```
artemmatiushenko1: `outerHeader`

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,56 @@
...
+                username={getFullName(user.firstName, user.lastName)}
+                avatarUrl={null}
+              />
+            </button>
+          </header>
+
+          <Popover
+            trigger={{ handleToggleModalOpen, isOpen }}
+            content={<DropDown trigger={{ handleToggleModalOpen, isOpen }} />}
+            className={`${styles.dropdown} ${styles.dropdownModal} ${
```
artemmatiushenko1: Please use `getValidClassnames` helper here

frontend/src/libs/components/link/link.tsx
```diff
@@ -32,7 +34,7 @@ const Link: React.FC<Properties> = ({
   );
 
   return (
-    <NavLink to={to} className={handleGetClassesByLinkState}>
+    <NavLink to={to} className={handleGetClassesByLinkState} onClick={onClick}>
```
artemmatiushenko1: I’m not sure that we should add `onClick` to link, maybe it will be better to use a `Button` in the place where you need that?

frontend/src/libs/components/header/header.tsx
kovAnya: Are you sure this is necessary? This className props gets to the Modal component, where getValidClassnames helper is already used. So classes are already correctly recognized.

andanf-e: Yep, use this helper for all classes composition

kovAnya: Thanks, got it

frontend/src/libs/components/header/header.tsx
```diff
@@ -1,30 +1,55 @@
...
+          </header>
+
+          <Popover
+            trigger={{ handleToggleModalOpen, isOpen }}
+            content={<Dropdown trigger={{ handleToggleModalOpen, isOpen }} />}
+            className={getValidClassNames(
+              styles.dropdown,
+              styles.dropdownModal,
+              isOpen ? styles.open : '',
+            )}
```
PablloZz, line 56: Why not just isOpen && styles.open?

frontend/src/libs/components/header/header.tsx
```diff
@@ -67,12 +52,15 @@ const Header: React.FC<Properties> = ({ user }) => {
               WRITORIUM
             </Link>
 
-            <div className={styles.avatarWrapper} ref={avatarReference}>
+            <button
+              className={styles.avatarButton}
+              onClick={handleClickOnAvatar}
+            >
```
PablloZz: Maybe it should be replaced with the custom Button component

kovAnya: button here is just a wrapper for Avatar component. I can't use a custom Button instead because it doesn't allow children

andanf-e (dismissed):

NazarKovtun1 (dismissed):

andanf-e (dismissed):

andanf-e (approved):

NazarKovtun1 (approved):

PR #142, opened 2023-09-04
wr-139: Add responsive for Sign In/Up pages

AnastasiaR2:
Added responsiveness for Auth pages. On the narrow screens, the right side with bg is hidden. Currently, the title is not responsive, so I disabled the font-size property to make a second screenshot. But I checked with the logo from Iryna's task, and it's working as expected. 

image_2023-09-04_12-51-02 
image_2023-09-04_12-51-56


andanf-e (commented):
Also show how forget password page looks

PR #143, opened 2023-09-04
wr-132:  Add date-fns package for dates

hnoievets:
- [x] Add [date-fns](https://date-fns.org/) package to shared folder
- [x]  Provide helpers for date formatting
- [x]  Replace places throughout the code to use date-fns helpers to manage dates shape


shared/src/libs/enums/date-pattern.emun.ts
```diff
@@ -0,0 +1,6 @@
+const DatePattern = {
```
NazarKovtun1: typo in file name

NazarKovtun1 (changes requested):

shared/src/libs/enums/date-pattern.emun.ts
hnoievets: Thanks

shared/src/libs/enums/date-pattern.enum.ts
```diff
@@ -0,0 +1,6 @@
+const DatePattern = {
+  MONTH_DAY: 'd',
+  SHORT_MONTH_NAME: 'LLL',
+} as const;
```
andanf-e:
```suggestion
const DateFormat = {
  DAY_SHORT_MONTH: 'd MMM',
} as const;
```

shared/src/libs/helpers/format-date/format-date.helper.ts
```diff
@@ -0,0 +1,23 @@
...
+    useAdditionalDayOfYearTokens?: boolean;
+  };
+};
+
+const formatDate = ({
+  isoString,
+  pattern,
+  options,
+}: FormatDateParameters): string => {
+  return format(new Date(isoString), pattern, options);
```
andanf-e:
```suggestion
const getFormattedDate = (date: string, format: ValueOf<typeof DateFormat>): string => {
  return format(new Date(date), format);
```

andanf-e (changes requested):

andanf-e (approved):

NazarKovtun1 (approved):

PR #154, opened 2023-09-06
wr-147: Display created Article with formatting

AnastasiaR2:
1. Implemented saving markup formatting on the article cards. 
2. Added styles for overlapped text to shorten it with '...'.
3. Updated some genre keys, so there are no conflicts with saving genre. 
'sci_fi' => 'science_fiction'
'hist_fiction'  => 'historical_fiction'
'ya_fiction'  => 'young_adult' 

image_2023-09-06_10-52-38


frontend/src/pages/articles/components/article-card/article-card.tsx
```diff
@@ -50,7 +54,10 @@ const ArticleCard: React.FC<Properties> = ({
       <div className={styles.body}>
         <div className={styles.articleInfo}>
           <h4 className={styles.title}>{title}</h4>
-          <p className={styles.text}>{text}</p>
+          <p
+            className={styles.text}
+            dangerouslySetInnerHTML={getDangerousHtmlObject(text)}
+          ></p>
```
andanf-e: Not sure if helper to get prop for `dangerouslySetInnerHTML` is necessary
```suggestion
          <article
            className={styles.text}
            dangerouslySetInnerHTML={{ __html: sanitizeHtml(htmlContent) }}
          />
```

frontend/src/pages/articles/components/article-card/styles.module.scss
```diff
@@ -72,14 +73,29 @@
...
+  --max-lines: 3;
+  --height: 51px;
+
+  display: -webkit-box;
+  -webkit-line-clamp: var(--max-lines);
+  -webkit-box-orient: vertical;
+  max-width: 585px;
+  height: var(--height);
+  margin: 14px 0 23px;
+  overflow: hidden;
```
andanf-e: Let's move text-overflow styling to `scaffolding` styles file as it's done for `visually-hidden` styling (we may also use text overflowing on overlap in other places, so it could be common)

andanf-e (changes requested):

frontend/src/pages/articles/components/article-card/styles.module.scss
AnastasiaR2: I think it will be good to create text-overflow styling for multiline and single-line (using for prompt tags, for example) in this case. Should I add `line-height: calc(var(--height) / var(--max-lines));` to text-overflow styling, so the texts aren't being cut off in case of fixed height?

andanf-e: `single/multi` line cases are more specific, so maybe let's leave them in specific styling

AnastasiaR2: So, in scaffolding styles, I should only add this class?
```
.text-overflow {
    overflow: hidden;
    text-overflow: ellipsis;
}
```

andanf-e: Yep

andanf-e (approved):

mhrechkin (approved):

PR #155, opened 2023-09-06
wr-146: link-prompt-to-articles

canterbery:
image


shared/src/packages/articles/libs/types/article-with-author.type.ts
```diff
@@ -4,6 +4,13 @@ import { type ArticleEntityType } from './article-entity.type.js';
 
 type ArticleWithAuthorType = ArticleEntityType & {
   author?: UserDetailsResponseDto;
+  prompt?: {
+    character: string;
+    setting: string;
+    situation: string;
+    prop: string;
+  };
```
andanf-e: I believe we already have `GeneratedArticlePrompt` type and it can be used both here and for BE model relation

frontend/src/pages/articles/libs/types/tag.type.ts
```diff
@@ -1,6 +1,7 @@
 type TagType = {
-  id: number;
-  name: string;
+  id: string;
+  category: string;
```
andanf-e: User `PromptCategory` enum values

frontend/src/pages/articles/libs/types/tag.type.ts
```diff
@@ -1,6 +1,7 @@
 type TagType = {
-  id: number;
-  name: string;
+  id: string;
```
andanf-e: I guess we don't need it
```suggestion

```

andanf-e (changes requested):

frontend/src/pages/articles/libs/helpers/get-article-tags.ts
```diff
@@ -0,0 +1,24 @@
...
+    });
+  }
+  if (article.prompt) {
+    for (const [key, value] of Object.entries(article.prompt)) {
+      if (value) {
+        result.push({ id: crypto.randomUUID(), category: key, text: value });
+      }
+    }
+  }
+  return result;
```
andanf-e: And disable rule restricting `.reduce` usage
```suggestion
  const result: TagType[] = Boolean(article.genre)
     ? [{ category: 'genre', text: article.genre }]
     : [];
  
  return Object.entries(article.prompt ?? {}).reduce((resultingTags, [key, value]) => {
    if (value) {
      return resultingTags.concat({ category: key, text: value });
    }
    return resultingTags
  }, result); 
```

andanf-e (changes requested):
And also change styling for tags a little bit - make block bg color as you've added for container + remove additional top-padding from bubble with category
image

shared/src/packages/articles/libs/types/article-with-author.type.ts
canterbery: In that type keys can be string only but I need string|null so I need to write another type , write another helper like WithNullableKeys or leave it as it is.

canterbery: GeneratedArticlePrompt has string type of keys. And I need string|null

frontend/src/pages/articles/libs/types/tag.type.ts
canterbery: There is no genre in prompt category , I will need to create separate type

frontend/src/pages/articles/libs/helpers/get-article-tags.ts
canterbery: type guard doesn't work with this variant 
image

andanf-e: Then just leave it as `article.genre` check

shared/src/packages/articles/libs/types/article-with-author.type.ts
andanf-e: What about `WithNullableKeys<GeneratedArticlePrompt>`

canterbery: It wants second argument for omit.
I need to write another type or helper or just leave it as it is.

andanf-e: I believe it may work `WithNullableKeys<GeneratedArticlePrompt, keyof GeneratedArticlePrompt>`

frontend/src/pages/articles/libs/helpers/get-article-tags.ts
```diff
@@ -0,0 +1,27 @@
...
+    let keys: keyof typeof prompt;
+    for (keys in prompt) {
+      const value = prompt[keys];
+      if (value) {
+        result.push({ category: keys, text: value });
+      }
+    }
+  }
+  return result;
+};
```
andanf-e, line 26: helper rework was not resolved

andanf-e (commented):
image
Also tags may look like this

frontend/src/pages/articles/libs/helpers/get-article-tags.ts
canterbery: Implemetation changed because Object.entries is no longer possible to use because of type issues.  That's why I submitted new variant and awaiting for review.  Also I beleive we agreed that genre check should remain as it is.

andanf-e: Fact that `key is string` on Object.entries does not mean that you have no ability to cast it to needed type - if you know that key actually would be of that type

canterbery: I thought you are not supposed to do it if possible because it defeats the purpose of typescript.

canterbery: Looks like we have some linter problems with this variant 
image

andanf-e: Yep, about no-array-reduce -- I've mentioned in previous comments that it may be disabled

andanf-e: About TS - you are using technology, this thesis should not work vice versa 🥲

canterbery: Well , it's a good point. I am just trying to understand why reduce is better here.
Should I disable prefer-spread rule as well ?

andanf-e: Spread can also be disabled, yes
Because with reduce you're taking new instance of entity without mutating it on any steps + reduce works with arrays that means we won't provide confusion with iterating through object (like it could be with for...in) or anything else

canterbery: Interesting , thank you.

mhrechkin (dismissed):

andanf-e (approved):

mhrechkin (approved):

PR #162, opened 2023-09-06
wr-153: make-email-case-insensitive

canterbery:


backend/src/db/migrations/20230906150722_make_users_email_index.ts
```diff
@@ -0,0 +1,13 @@
+import { type Knex } from 'knex';
+
+async function up(knex: Knex): Promise<void> {
+  return await knex.raw(
+    'CREATE UNIQUE INDEX users_lowercase_email ON users (lower(email));',
+  );
+}
+
+async function down(knex: Knex): Promise<void> {
+  return await knex.raw('DROP index IF EXISTS users_lowercase_email');
```
andanf-e: Initially it has just unique constraint, so I believe it also should be moved back on rollback

backend/src/packages/users/user.repository.ts
```diff
@@ -36,7 +36,7 @@ class UserRepository implements IRepository {
   public async findByEmail(email: string): Promise<UserEntity | null> {
     const user = await this.userModel
       .query()
-      .where({ email })
+      .whereRaw('LOWER(email)=?', [email.toLowerCase()])
```
andanf-e: As alternative
```suggestion
      .where('email', 'ilike', email)
```

backend/src/db/migrations/20230906150722_make_users_email_index.ts
```diff
@@ -0,0 +1,13 @@
+import { type Knex } from 'knex';
+
+async function up(knex: Knex): Promise<void> {
+  return await knex.raw(
+    'CREATE UNIQUE INDEX users_lowercase_email ON users (lower(email));',
```
andanf-e: Let's also drop default constraint if it exists
`users_email_unique`

andanf-e (changes requested):

backend/src/db/migrations/20230906150722_make_users_email_index.ts
canterbery: I think DROP INDEX also removes UNIQUE constraint if I understood you correctly. 
image

andanf-e: I mean drop initial unique constraint on migration,, and turn it back on migration rollback

NazarKovtun1 (dismissed):

backend/src/db/migrations/20230906150722_make_users_email_index.ts
```diff
@@ -0,0 +1,15 @@
+import { type Knex } from 'knex';
+
+async function up(knex: Knex): Promise<void> {
+  return await knex.raw(
+    'CREATE UNIQUE INDEX IF NOT EXISTS users_lowercase_email ON users (lower(email)); alter table users drop constraint users_email_unique;',
+  );
```
andanf-e:
```suggestion
  await knex.raw(
    'CREATE UNIQUE INDEX IF NOT EXISTS users_lowercase_email ON users (LOWER(email));'
  );
  await knex.raw(
    'ALTER TABLE users DROP CONSTRAINT users_email_unique;'
  );
```

backend/src/db/migrations/20230906150722_make_users_email_index.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+async function down(knex: Knex): Promise<void> {
+  return await knex.raw(
+    'DROP index IF EXISTS users_lowercase_email; ALTER TABLE users ADD CONSTRAINT user_email_unique UNIQUE (email);',
+  );
```
andanf-e: Just to make queries more readable
```suggestion
  await knex.raw(
    'DROP INDEX IF EXISTS users_lowercase_email;'
  );
  await knex.raw(
    'ALTER TABLE users ADD CONSTRAINT user_email_unique UNIQUE (email);',
  );
```

mhrechkin (approved):

andanf-e (approved):

PR #174, opened 2023-09-07
wr-15: fix the header on top of the page during scroll

PablloZz:
image


frontend/src/libs/components/app/styles.module.scss
```diff
@@ -1,4 +1,5 @@
 .container {
   flex-grow: 1;
   width: 100%;
+  margin-top: 105px;
```
andanf-e: Let's move header height to variables and use it both for top-margin and for fixing header's height - then we'd be confident with this values

andanf-e (dismissed):

frontend/src/libs/components/header/header.tsx
```diff
@@ -23,7 +23,7 @@ const Header: React.FC<Properties> = ({ user }) => {
   return (
     <>
       {user && (
-        <div>
+        <div className={styles.container}>
```
drMangust: I think we can get rid of this `div` tag.

frontend/src/libs/components/header/styles.module.scss
```diff
@@ -1,8 +1,17 @@
...
+
 .header {
   position: relative;
   display: flex;
   justify-content: space-between;
   align-items: center;
+  height: var(--header-height);
   padding: 30px 60px;
   background: var(--light-grey);
   border-bottom: 2px solid var(--light-grey-15);
```
drMangust: get rid of `div` tag and container class.
and it will be enough to add styles to header

```suggestion
.header {
  position: fixed;
  z-index: var(--z-index-level-one);
  display: flex;
  justify-content: space-between;
  align-items: center;
  width: 100%;
  height: var(--header-height);
  padding: 30px 60px;
  background: var(--light-grey);
  border-bottom: 2px solid var(--light-grey-15);
}
```

andanf-e (approved):

frontend/src/assets/css/variables.scss
```diff
@@ -28,11 +28,15 @@
   --picasso: #ffec99;
   --madang: #b2f2bb;
   --overlay-alpha-50: rgb(0 0 0 / 50%);
+  --pale-grey: #dbdbdb;
```
IrynaK-a, line 31: we already have this variable: --light-grey-border: #dbdbdb;

mhrechkin (approved):

frontend/src/assets/css/variables.scss
PablloZz: This color wasn't added by me, I just put that color back because Volodymyr accidentally deleted it and styles fell of

PR #177, opened 2023-09-07
wr-9: Implement logout after auth token expired

IrynaK-a:


frontend/src/libs/packages/api/http-api.package.ts
```diff
@@ -105,6 +105,11 @@ class HttpApi implements IHttpApi {
       }),
     )) as ServerErrorResponse;
 
+    if (response.status === 401) {
```
andanf-e:
```suggestion
    if (response.status === HttpCode.UNAUTHORIZED) {
```

frontend/src/libs/packages/api/http-api.package.ts
```diff
@@ -105,6 +105,11 @@ class HttpApi implements IHttpApi {
...
+      await this.storage.drop(StorageKey.TOKEN);
+      window.location.assign('/sign-in');
```
andanf-e:
```suggestion
      window.location.assign(AppRoute.SIGN_IN);
```

andanf-e (changes requested):

mhrechkin (dismissed):

frontend/src/libs/packages/api/http-api.package.ts
```diff
@@ -105,6 +105,11 @@ class HttpApi implements IHttpApi {
       }),
     )) as ServerErrorResponse;
 
+    if (response.status === HttpCode.UNAUTHORIZED) {
```
andanf-e: We may also make it more correct way - like applying `jwt-decode` package to retrieve exp_time and check it.
Because I thought a little bit: and with this approach any user action that fires UNAUTHORIZED - will trigger logout

andanf-e: As alternative to decoding token and checking it's exp_time -- we may just check additional response error message if it's equal to `Invalid Token`

frontend/src/libs/packages/api/http-api.package.ts
```diff
@@ -105,6 +105,14 @@ class HttpApi implements IHttpApi {
       }),
     )) as ServerErrorResponse;
 
+    if (
+      response.status === HttpCode.UNAUTHORIZED &&
+      parsedException.message === 'Invalid token'
```
andanf-e: As you're using save ExceptionMessage as from BE - then move it to shared

andanf-e (changes requested):

andanf-e (dismissed):

mhrechkin (dismissed):

andanf-e (approved):

PR #180, opened 2023-09-07
wr-146: fix-null-prompts-display

canterbery:


backend/src/packages/articles/article.repository.ts
```diff
@@ -35,12 +35,14 @@ class ArticleRepository implements IRepository {
...
-          prop: article.prompt.prop,
-        },
+        prompt: article.prompt
+          ? {
+              character: article.prompt.character,
+              setting: article.prompt.setting,
+              situation: article.prompt.situation,
+              prop: article.prompt.prop,
+            }
+          : undefined,
```
andanf-e: Pass `null` explicitly instead of undefined

andanf-e (changes requested):

backend/src/packages/articles/article.repository.ts
canterbery: It's an optional parameter that's why undefined.  In order to pass null I will have to change type and rewrite all initialize methods in article entity

mhrechkin (approved):

andanf-e (approved):

PR #186, opened 2023-09-09
wr-172: Show user uploaded avatar in users header circle

AnastasiaR2:
- If a user uploads the avatar, it will be shown in the header, article card, and article page. 
- Fixed bug with the Read more redirect link.

image_2023-09-09_09-33-24
image_2023-09-09_09-33-44
image_2023-09-09_09-34-05


frontend/src/pages/article/article-page.tsx
```diff
@@ -19,8 +19,12 @@ const ArticlePage: React.FC = () => {
 
   const { id } = useParams();
 
+  const [initialDataLoad, setInitialDataLoad] = useState(false);
```
AnastasiaR2: I found that dataStatus is initially fulfilled here because the state is updated on the previous page when all articles are loaded. That's why, there was a bug with the Redirect More link (an immediate redirect occurred before the data was loaded). To fix it, I added a separate state variable to track whether the initial data loaded.

andanf-e: I believe we don't need this boolean flag, let's just try to fix `dataStatus` management in article slice -- for now - just create separate dataStatus for getArticle action

shared/src/packages/users/libs/types/user-details-dto.type.ts
```diff
@@ -2,6 +2,10 @@ type UserDetailsDto = {
   id: number;
   firstName: string;
   lastName: string;
+  avatarUrl: string | null;
+  avatar?: {
+    url: string | null;
+  };
```
andanf-e: Why do we need to duplicate it? Can't we just have a single avatarUrl field?

frontend/src/pages/article/article-page.tsx
```diff
@@ -60,6 +64,7 @@ const ArticlePage: React.FC = () => {
           {author && (
             <AuthorDetails
               name={getFullName(author.firstName, author.lastName)}
+              avatarUrl={author.avatarUrl}
```
andanf-e: Please also pass actual article's `publishedAt` and `genre`

frontend/src/pages/article/components/author-details/author-details.tsx
```diff
@@ -22,12 +23,13 @@ const AuthorDetails: FC<Properties> = ({
   publishedAt = '2023-09-04T12:53:07.144Z',
   readingTime = '7 min',
   genre = 'Fiction',
+  avatarUrl = null,
```
andanf-e: Please remove default value assignment from `publishedAt` and `genre` and also remove optionality from props type (for genre I believe as of now type of prop should be `string | null`

andanf-e (changes requested):

mhrechkin (dismissed):

shared/src/packages/users/libs/types/user-details-dto.type.ts
AnastasiaR2: If I remove
```
 avatar?: {
    url: string | null;
  };
```
I will get an error that the property 'avatar' does not exist on type 'UserDetailsResponseDto' in the article entity. Is there a workaround for this?

andanf-e: Not a workaround, but declaring a proper type for entity

shared/src/packages/users/libs/types/user-details-dto.type.ts
```diff
@@ -2,6 +2,9 @@ type UserDetailsDto = {
   id: number;
   firstName: string;
   lastName: string;
+  avatar: {
+    url: string | null;
+  };
```
andanf-e:
```suggestion
  avatarUrl: string | null;
```

frontend/src/slices/articles/articles.slice.ts
```diff
@@ -33,7 +35,7 @@ const { reducer, actions, name } = createSlice({
       state.dataStatus = DataStatus.FULFILLED;
     });
     builder.addCase(getArticle.fulfilled, (state, action) => {
-      state.dataStatus = DataStatus.FULFILLED;
+      state.getArticleStatus = DataStatus.FULFILLED;
```
andanf-e, line 56: I guess now it should also be used for `getArticle.pending` and `getArticle.rejected`

andanf-e (changes requested):

shared/src/packages/articles/libs/types/article-with-author.type.ts
```diff
@@ -3,7 +3,7 @@ import { type UserDetailsResponseDto } from '~/packages/users/users.js';
 import { type ArticleEntityType } from './article-entity.type.js';
 
 type ArticleWithAuthorType = ArticleEntityType & {
-  author?: UserDetailsResponseDto;
+  author?: UserDetailsResponseDto | null;
```
andanf-e: I guess it can't be null, as article is always created by user

frontend/src/pages/article/article-page.tsx
```diff
@@ -48,7 +48,7 @@ const ArticlePage: React.FC = () => {
     { id: 5, name: 'Tech' },
   ];
 
-  const { text, title, author, coverUrl } = article ?? {};
+  const { text, title, author, coverUrl, publishedAt, genre } = article ?? {};
```
andanf-e: Not sure why do you need nullish coalescing - as you've already checked that `article` exists

AnastasiaR2: Seems that the data is not fully loaded at the point of destructuring, and if nullish coalescing is removed, I get an error:
Cannot destructure property 'text' of 'article' as it is null.
I tried to set `isLoading to getArticleStatus === DataStatus.PENDING`. It helped, but the issue with read more redirect returns in this case.

AnastasiaR2: May I leave a default value of empty {} to prevent the error?

andanf-e (approved):

NazarKovtun1 (approved):

PR #195, opened 2023-09-10
wr-193: add read time property to article

artemmatiushenko1:
This pull request adds a read time property to article entity.
- Added `read_time` column to articles table
- Updated article entity/model/repository to support new property
- Get read time on article create/update using openai service
- Display real `readTime` value on article details page and article card

image
image


backend/src/db/migrations/20230910082133_add_article_read_time_column.ts
```diff
@@ -0,0 +1,19 @@
+import { type Knex } from 'knex';
+
+const TableName = 'articles' as const;
+
+const ColumnName = 'read_time' as const;
```
andanf-e: Primitive constants are already constants
```suggestion
const TABLE_NAME = 'articles';

const COLUMN_NAME = 'read_time';
```

andanf-e (dismissed):

NazarKovtun1 (dismissed):

andanf-e (approved):

PR #201, opened 2023-09-11
wr-15: Fix header margin on auth pages

PablloZz:
image
image


andanf-e (approved):

frontend/src/assets/css/variables.scss
```diff
@@ -36,9 +36,6 @@
   --z-index-level-one: 100;
   --z-index-tooltip: 1000;
   --achievement-columns: 3;
-
-  /* header height */
-  --header-height: 105px;
```
IrynaK-a, line 41: @PablloZz @andanf-e Is it a good idea to delete this variable?

andanf-e: If it won't be used, then yes

PR #205, opened 2023-09-12
wr-137: Show user activity in profile

PablloZz:
Fixed incorrect alignment of days
image
image


frontend/src/pages/profile/components/user-activity/user-activity.tsx
```diff
@@ -13,18 +18,27 @@ type Properties = {
...
     return (
       <span
-        title={date}
-        key={date}
+        title={activityTitle}
+        key={localDate}
         className={getValidClassNames(
           styles.activityItem,
           styles[activityBreakpoint],
+          index === 0 &&
```
andanf-e: **B31**

frontend/src/pages/profile/components/user-activity/user-activity.tsx
```diff
@@ -13,18 +18,27 @@ type Properties = {
 
 const UserActivity: React.FC<Properties> = ({ userActivity }) => {
   const uniqueMonths: string[] = getUniqueMonths(userActivity);
+  const dateFormatter = new Intl.DateTimeFormat(DATE_LOCALE, DATE_OPTIONS);
 
-  const activityStatistic = userActivity.map((activity) => {
+  const activityStatistic = userActivity.map((activity, index) => {
     const { count, date } = activity;
     const activityBreakpoint = getActivityBreakpoint(count);
+    const localDate = dateFormatter.format(new Date(date));
```
andanf-e: Let's use getFormattedDate helper, as it's already added

frontend/src/pages/profile/components/user-activity/user-activity.tsx
```diff
@@ -13,18 +18,27 @@ type Properties = {
...
+    const activityTitle = `${count} action${
+      count > 1 ? 's' : ''
+    } on ${localDate}`;
```
andanf-e: I believe in `getPublicationTime` helper - there is already the same check to add pluralEnding -- let's move it to separate helper and use in both places

andanf-e (changes requested):

frontend/src/pages/profile/components/user-activity/user-activity.tsx
PablloZz: Is there function to convert date to the current timezone?
Because from server dates come one day shifted

PablloZz: I tried to find such a method there but it seems like it hasn't a method to automatically convert a date to the current timezone
This is necessary because the dates from the server are shifted by one day

andanf-e: `format` method in date-fns has separate `options` param - couldn't utc cases be resolved by it?

PablloZz: Oh, yes, we could, at first I just didn't want to use enUs as the locale and wanted the locale to be defined automatically but I changed my mind when I saw Ukrainian words in the date )

andanf-e (approved):

NazarKovtun1 (approved):

PR #210, opened 2023-09-12
wr-199: Add ability to Delete Articles

AnastasiaR2:
1. Added Delete button for the article on my-articles page.
2. On Delete article action made soft-delete (added deleted_at col and when the article is chosen for removal - it's populated with the current date) of the chosen article.
3. Added default filtration on articles to hide deleted ones (whereNull('deletedAt')).

image


frontend/src/pages/articles/components/article-card/article-card.tsx
```diff
@@ -74,12 +81,29 @@ const ArticleCard: React.FC<Properties> = ({
...
+            <>
+              <IconButton
+                className={styles.iconButton}
+                iconName="trashBin"
+                iconClassName={getValidClassNames(
+                  styles.deleteIcon,
+                  styles.pointerIcon,
+                )}
+                onClick={handleDeleteArticle}
+              ></IconButton>
```
artemmatiushenko1: Please use self-closing tag
```jsx
    <IconButton
      className={styles.iconButton}
      iconName="trashBin"
      iconClassName={getValidClassNames(
        styles.deleteIcon,
        styles.pointerIcon,
      )}
      onClick={handleDeleteArticle}
    />
```

frontend/src/packages/articles/articles-api.package.ts
```diff
@@ -101,6 +101,19 @@ class ArticleApi extends HttpApi {
 
     return await response.json<ArticleWithAuthorType>();
   }
+
+  public async delete(id: number): Promise<ArticleWithAuthorType> {
+    const response = await this.load(
+      this.getFullEndpoint(ArticlesApiPath.$ID, { id: String(id) }),
+      {
+        method: 'DELETE',
+        contentType: ContentType.FORM_DATA,
```
artemmatiushenko1: Why `contentType` is `FORM_DATA`? Shouldn't it be `ContentType.JSON` 🤔

backend/src/db/migrations/20230911133910_add_deleted_at_column_to_articles.ts
```diff
@@ -0,0 +1,21 @@
+import { type Knex } from 'knex';
+
+const TableName = 'articles';
```
artemmatiushenko1: `const TABLE_NAME = 'articles';`

backend/src/db/migrations/20230911133910_add_deleted_at_column_to_articles.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const ColumnName = {
```
artemmatiushenko1: `const COLUMN_NAME = 'deleted_at'`

andanf-e (commented):
Waiting on comments to be resolved

frontend/src/packages/articles/articles-api.package.ts
AnastasiaR2: Actually, missed the payload field, and there was an error with ContentType.JSON. But now resolved, thanks)

backend/src/db/migrations/20230913044708_add_deleted_at_column_to_articles.ts
```diff
@@ -0,0 +1,21 @@
+import { type Knex } from 'knex';
+
+const TABLE_NAME = 'articles';
+
+const COLUMN_NAME = {
+  DELETED_AT: 'deleted_at',
+} as const;
```
andanf-e: I guess Artem meant that in case if you have one value in enum - then you don't need a enum, but could just have a constant
```suggestion
const DELETED_AT = 'deleted_at',;
```

frontend/src/packages/articles/articles-api.package.ts
```diff
@@ -101,6 +101,20 @@ class ArticleApi extends HttpApi {
...
     return await response.json<ArticleWithAuthorType>();
   }
+
+  public async delete(id: number): Promise<ArticleWithAuthorType> {
+    const response = await this.load(
+      this.getFullEndpoint(ArticlesApiPath.$ID, { id: String(id) }),
+      {
+        method: 'DELETE',
+        contentType: ContentType.JSON,
+        payload: JSON.stringify(id),
```
andanf-e: You have already passed id as param + taking it from params at BE
```suggestion
```

andanf-e (changes requested):

frontend/src/packages/articles/articles-api.package.ts
AnastasiaR2: If I leave just `contentType: ContentType.JSON`, I will get an error: Body cannot be empty when content-type is set to 'application/json'. I cannot also remove contentType as it's required in HttpApiOptions. I thought of using PATCH instead of DELETE, cause we are actually changing deleteAt field, but there is already a PATCH method on the articles/id route. How I can resolve this? May I make contentType optional?

andanf-e: Then why can't you just make `contentType` optional? and in getHeaders attach it only if it's present

AnastasiaR2: Got it, will do

andanf-e (dismissed):

andanf-e (approved):

NazarKovtun1 (approved):

PR #212, opened 2023-09-12
wr-209:  Make hoverable tooltips style identical on Prompt tags and Achievement tab

hnoievets:
- [x]  Make the hoverable tooltips on Prompt tags the same as on the Achievement tab
- [x]  The background color of the tooltips should be #2E453B (both on Prompt tags and the Achievement tab)

Знімок екрана 2023-09-13 003509
Знімок екрана 2023-09-13 003526


andanf-e (dismissed):

frontend/src/pages/profile/components/user-achievements/styles.module.scss
```diff
@@ -43,6 +43,7 @@
 .achievementBlock .achievementTooltip {
   max-width: 150px;
   color: var(--pale-gray);
-  background-color: var(--light-green);
+  background-color: var(--dark-green);
   border-radius: 7px;
+  opacity: 100;
```
artemmatiushenko1: I believe it should be `opacity: 1`

andanf-e (dismissed):

mhrechkin (approved):

andanf-e (approved):

Issue #136, opened 2023-09-04
Show achievements on Profile


andanf-e:
![image](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/assets/19575839/fa256f25-3a4e-4568-8078-8a96cb0bada4)

- [x] Create component to check user achievements
- [x] For now - just show few random achievements as user's achieved
- [x] Use modal to show whole list of achievements
- [x] As of now we'd not have proper design for it - so use simple view as from design mockup

lera777:
- [x] The tooltip should be located within the modal window.


**Environment:**
OS: macOS Catalina 10.15.7
Browser: Google Chrome 116.0.5845.110

**Severity:**
Low

**Priority:**
Medium

PR #219, opened 2023-09-13
wr-136: Fix Show achievements on Profile

drMangust:
task [Bug fix] - [Show achievements on Profile#136](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/issues/136#issuecomment-1712442669)

- [x]  The tooltip should be located within the modal window.
- [x] add tooltip position strategy

image
image


andanf-e (approved):

frontend/src/pages/profile/components/user-achievements/styles.module.scss
```diff
@@ -28,6 +32,7 @@
 }
 
 .modal {
+  overflow: hidden;
```
IrynaK-a, line 35: Do we need this? The Modal component has it's style `overflow: auto;`

drMangust: In this case it is crucial for us that tooltip should be located within the modal window.
This is required for tooltip-position-strategy
https://react-tooltip.com/docs/options

Examples:

image
image

IrynaK-a: Okay, I got it. Thanks!

NazarKovtun1 (approved):

PR #247, opened 2023-09-16
wr-200: Join Article actions to popover

AnastasiaR2:
1. Wrapped Article action buttons in the tooltip
2. Changed tooltip component, so it can be opened on click, added handling of outside click
3. Fixed issue with the Read More link

image_2023-09-16_09-28-09
image_2023-09-16_09-28-46


frontend/src/libs/components/tooltip/tooltip-wrapper.tsx
```diff
@@ -3,31 +3,92 @@ import 'react-tooltip/dist/react-tooltip.css';
...
+    (event: MouseEvent): void => {
+      if (
+        tooltipReference.current &&
+        !tooltipReference.current.contains(event.target as Node)
+      ) {
+        onClose && onClose();
+      }
+    },
+    [onClose],
+  );
```
AnastasiaR2: Currently, there is an issue with disabling scroll. When clicking outside the tooltip and it's a click on another more actions button, two handlers are triggering: click on the button and the outside click. And the state of isOpen will be true, and then false. The tooltip is visible, but in this case, the document scroll isn't disabled as it's related to the isOpen state. Could you suggest some solution here?
Meanwhile, adding a ref to the ReactTooltip itself doesn't work, I added it to the anchorElement, which is a more actions button in this case.

frontend/src/libs/components/icon/common.ts
```diff
@@ -125,6 +127,7 @@ const iconNameToIcon: Record<
   header6: Header6,
   trashBin: TrashBinIcon,
   facebook: Facebook,
+  threeDotsVertical: ThreeDotsVertical,
```
artemmatiushenko1: Maybe it would be better to name it as `ellipsisVertical`

frontend/src/libs/components/tooltip/tooltip-wrapper.tsx
artemmatiushenko1: I'm not sure if it is a good idea to use `ReactTooltip` for popovers at all because tooltips have been designed for quite different purposes, so it might produce unexpected behavior (like this one) if we use them differently. Maybe we should take a look at some other packages (Radix UI Popover for example). It's a headless composable component, with all accessibility and functionality in place.
@andanf-e what do you think?

https://www.radix-ui.com/primitives/docs/components/popover
https://www.npmjs.com/package/@radix-ui/react-popover

andanf-e: Let's leave it as it is, I will take a look on it by myself

frontend/src/libs/components/tooltip/tooltip-wrapper.tsx
```diff
@@ -3,31 +3,92 @@ import 'react-tooltip/dist/react-tooltip.css';
...
+}) => {
+  const tooltipReference = useRef<HTMLDivElement>(null);
+
+  const handleClickOutside = useCallback(
+    (event: MouseEvent): void => {
+      if (
+        tooltipReference.current &&
+        !tooltipReference.current.contains(event.target as Node)
+      ) {
+        onClose && onClose();
```
PablloZz: Maybe let's just use onClose?.()?

andanf-e (commented):
Please check if Popover component is now suitable

frontend/src/libs/components/tooltip/tooltip-wrapper.tsx
AnastasiaR2:
```
useEffect(() => {
    if (isOpen && !isScrollable) {
      document.body.classList.add(styles.noOverflow);
    }

    return () => {
      document.body.classList.remove(styles.noOverflow);
    };
  }, [isOpen, isScrollable]);
```
I added this code to the popover component to disable scroll when the popover is opened. But the issue with the scroll is actually the same as it was before. Should I handle it differently or just omit the scroll functionality for now? Also, as there is a popover component, do I need to remove modifications in the tooltip component?

andanf-e: Yep, left it as it is with scrolling,
And yes, remove changes from tooltip

frontend/src/assets/img/icons/ellipsis-vertical-icon.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" viewBox="0 -960 960 960"  fill="currentColor"><path d="M480-160q-33 0-56.5-23.5T400-240q0-33 23.5-56.5T480-320q33 0 56.5 23.5T560-240q0 33-23.5 56.5T480-160Zm0-240q-33 0-56.5-23.5T400-480q0-33 23.5-56.5T480-560q33 0 56.5 23.5T560-480q0 33-23.5 56.5T480-400Zm0-240q-33 0-56.5-23.5T400-720q0-33 23.5-56.5T480-800q33 0 56.5 23.5T560-720q0 33-23.5 56.5T480-640Z" fill="#44996B"/></svg>
```
andanf-e: missing newline ending

frontend/src/pages/articles/components/article-card/libs/components/popover-buttons-group/popover-buttons-group.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <>
+            <Icon iconName="favorite" />
+            <span>Add to favorites</span>
+          </>
+        }
+      />
+      {isOwnArticle && (
+        <>
+          <RouterLink
+            to={AppRoute.ARTICLES_EDIT_$ID.replace(':id', id.toString())}
```
andanf-e: Use `configureString` for it

andanf-e (approved):

andanf-e (changes requested):
Sorry, approved by accident

frontend/src/pages/articles/components/article-card/libs/components/popover-buttons-group/popover-buttons-group.tsx
```diff
@@ -0,0 +1,73 @@
...
+        label={
+          <>
+            <Icon iconName="favorite" />
+            <span>Add to favorites</span>
+          </>
+        }
+      />
+      {isOwnArticle && (
+        <>
+          <RouterLink
```
andanf-e: Use Link common component

andanf-e (changes requested):

frontend/src/pages/articles/components/article-card/libs/components/popover-buttons-group/popover-buttons-group.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <div className={styles.buttonsGroup}>
+      <Button
+        label={
+          <>
+            <Icon iconName="favorite" />
+            <span>Add to favorites</span>
+          </>
+        }
+      />
```
andanf-e: There is IconButton component - can't it be used here?

frontend/src/pages/articles/components/article-card/libs/components/popover-buttons-group/popover-buttons-group.tsx
```diff
@@ -0,0 +1,71 @@
...
+            className={styles.editLink}
+          >
+            <Button
+              label={
+                <>
+                  <Icon iconName="edit" />
+                  <span>Edit</span>
+                </>
+              }
+            />
```
andanf-e: I'm not sure you need this button inside link, when you can just style link same way as buttons in popover

frontend/src/pages/articles/components/article-card/libs/components/popover-buttons-group/popover-buttons-group.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </Link>
+          <Button
+            label={
+              <>
+                <Icon iconName="trashBin" />
+                <span>Delete</span>
+              </>
+            }
+            onClick={handleDeleteArticle}
+          />
```
andanf-e: Same about IconButton

andanf-e (changes requested):

andanf-e (approved):

mhrechkin (approved):

PR #253, opened 2023-09-17
wr-90: assign unknown genre to article in case openai fails to detect it

artemmatiushenko1:


backend/src/db/migrations/20230917152623_add_unknown_genre.ts
```diff
@@ -0,0 +1,23 @@
...
+    .where(GENRE_KEY_COLUMN_NAME, UNKNOWN_GENRE.key)
+    .then(async (rows) => {
+      if (!rows.length) {
+        await knex(TABLE_NAME).insert(UNKNOWN_GENRE);
+      }
+    });
+};
+
+const down = (knex: Knex): Promise<void> => {
+  return knex(TABLE_NAME).where({ key: UNKNOWN_GENRE.key }).del();
```
andanf-e: If we have condition to insert `unknown` genre (when it does not exists), then we should not remove it from table on rollback

backend/src/db/migrations/20230917152623_add_unknown_genre.ts
```diff
@@ -0,0 +1,23 @@
...
+const UNKNOWN_GENRE = { key: 'unknown', name: 'Unknown' } as const;
+
+const up = async (knex: Knex): Promise<void> => {
+  return await knex(TABLE_NAME)
+    .where(GENRE_KEY_COLUMN_NAME, UNKNOWN_GENRE.key)
+    .then(async (rows) => {
+      if (!rows.length) {
+        await knex(TABLE_NAME).insert(UNKNOWN_GENRE);
+      }
+    });
```
andanf-e: I believe we can try to do it this way
```suggestion
    .insert(UNKNOWN_GENRE)
    .onConlict(GENRE_KEY_COLUMN_NAME)
    .ignore();
```

andanf-e (dismissed):

mhrechkin (approved):

andanf-e (approved):

PR #254, opened 2023-09-17
wr-235: Add scroll to top

kovAnya:


andanf-e (commented):
I'm not sure we should add scroll to top on each page instead of checking the root cause of scroll position stuck in place from previous page

andanf-e (approved):

NazarKovtun1 (approved):

PR #258, opened 2023-09-18
wr-137: Show user activity section in profile

PablloZz:
image

### DIFF
--- a/frontend/src/pages/profile/components/user-activity/styles.module.scss
+++ b/frontend/src/pages/profile/components/user-activity/styles.module.scss
@@ -7,6 +7,7 @@
   justify-content: center;
   align-items: center;
   padding: 20px 40px;
+  overflow-x: auto;
 }
 
 .title {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/profile/components/user-activity/styles.module.scss
```diff
@@ -15,7 +16,7 @@
 
 .activityWrapper {
   display: flex;
-  max-width: 710px;
+  width: 710px;
```
andanf-e, line 19: Not sure if we need to lock sections width directly

PablloZz: With max-width, the scroll doesn't appear in the firefox and table collapses

andanf-e: Okay, let's take a look

andanf-e (approved):

NazarKovtun1 (approved):

PR #277, opened 2023-09-19
wr-234: add ability to change article font size

canterbery:
image


frontend/src/libs/components/text-editor/libs/enums/text-size.ts
andanf-e, line 1: Files with enums should have .enum suffix (eg `text-size.enum`). I guess I've missed that part for other enums for this entity 😢

frontend/src/libs/components/text-editor/libs/extensions/font-size/libs/helpers/convert-font-size-to-number.helper.ts
```diff
@@ -0,0 +1,20 @@
...
+  styleValue: string | null | undefined,
+): number | null => {
+  if (!styleValue) {
+    return null;
+  }
+  const matches = styleValue.match(FONT_SIZE_PATTERN);
+  if (!matches) {
+    return null;
+  }
+  const value = Number(matches[1]);
```
andanf-e:
```suggestion
  const [, fontSizeDigit] = matches;
  const value = Number(fontSizeDigit);
```

frontend/src/libs/components/text-editor/libs/extensions/font-size/libs/helpers/increment-font-size.helper.ts
```diff
@@ -0,0 +1,10 @@
+import { FontSizeConfig } from '../constants/config.constant.js';
+
+const incrementFontSize = (fontSize: number): number => {
+  const newFontSize = fontSize + FontSizeConfig.CHANGE_STEP;
+  return newFontSize > FontSizeConfig.MAX_FONT_SIZE
+    ? FontSizeConfig.MAX_FONT_SIZE
+    : newFontSize;
```
andanf-e, line 7: I believe the same, but more compact
```suggestion
  return fontSize === FontSizeConfig.MAX_FONT_SIZE
    ? fontSize
    : fontSize + FontSizeConfig.CHANGE_STEP;
```

canterbery: But what if starting font size is an odd number ? Let's say 21px.  So it will go to 23 on first increment and then to 25 on second ( and our limit is 24 )

andanf-e: Okay 🤔

andanf-e (approved):

NazarKovtun1 (approved):

PR #284, opened 2023-09-20
wr-237: add genre check for articles with prompt

canterbery:


backend/src/packages/articles/article.service.ts
```diff
@@ -135,6 +136,63 @@ class ArticleService implements IService {
     return newGenreEntity.toObject().id;
   }
 
+  private async genreCheck(
```
andanf-e:
```suggestion
  private async checkActualGenre(
```

backend/src/packages/articles/article.service.ts
```diff
@@ -135,6 +136,63 @@ class ArticleService implements IService {
...
+
+    if (!genresJSON) {
+      return genreId;
+    }
+
+    const parsedGenres = safeJSONParse<DetectedArticleGenre[]>(genresJSON);
+    if (
+      !parsedGenres ||
+      !Array.isArray(parsedGenres) ||
+      parsedGenres.length === 0
```
andanf-e:
```suggestion
      !parsedGenres.length
```

backend/src/packages/articles/article.service.ts
```diff
@@ -135,6 +136,63 @@ class ArticleService implements IService {
...
+
+    const currentGenreInSuggestions =
+      parsedGenres.some((item) => item.key === currentGenre.toObject().key) &&
+      currentGenre.toObject().key !== UNKNOWN_GENRE_KEY;
+
+    if (currentGenreInSuggestions) {
+      return genreId;
+    }
+
+    const suggestedGenre = parsedGenres[0];
```
andanf-e:
```suggestion
    const [suggestedGenre] = parsedGenres;
```

backend/src/packages/articles/article.service.ts
```diff
@@ -342,6 +404,12 @@ class ArticleService implements IService {
       throw new ForbiddenError('Article can be edited only by author!');
     }
 
+    let updatedGenreId = payload.genreId ?? article.genreId;
+
+    if (payload.text && (article.promptId || payload.promptId)) {
```
andanf-e, line 435:
```suggestion
    const hasToCheckActualGenre = payload.text && (article.promptId || payload.promptId);
    if (hasToCheckActualGenre) {
```

backend/src/packages/articles/libs/helpers/get-detect-article-genre-completion-config.helper.ts
```diff
@@ -4,7 +4,7 @@ import { UNKNOWN_GENRE_KEY } from '~/packages/genres/genre.js';
 const getDetectArticleGenreCompletionConfig = (
   text: string,
 ): CompletionConfig => ({
-  prompt: `You will be provided with a html markup of an article, and your task is to extract a list of possible genres of that article. Provide your output in json format as an array of objects with keys name and key, where name is a readble name of the genre and key - a short snake cased representation of genre name. Place the most relevant genre at the 0 index. In case you can't indentify the genre put an object { key: "${UNKNOWN_GENRE_KEY}", name: "Unknown" } as the first item in resulting array. Here is the text: ${text}.`,
+  prompt: `You will be provided with a html markup of an article, and your task is to extract a list of possible genres of that article. Provide your output in json format as an array of objects with keys name and key, where name is a readble name of the genre and key - a short snake cased representation of genre name. Place the most relevant genre at the 0 index. In case you can't indentify the genre put an object { key: "${UNKNOWN_GENRE_KEY}", name: "Unknown" } as the last item in resulting array. Here is the text: ${text}.`,
```
artemmatiushenko1, line 7: When I was testing the prompt it was working kinda fine, but as it was discussed in Slack, for some reason currently ai misunderstands this part `In case you can't identify the genre put an object { key: "${UNKNOWN_GENRE_KEY}", name: "Unknown" } as the last item in resulting array`. Maybe we should remove it at all? 🤔

backend/src/packages/articles/article.service.ts
canterbery: you will lose payload.text type guard this way and will have to check it again

backend/src/packages/articles/libs/helpers/get-detect-article-genre-completion-config.helper.ts
canterbery: I tried to delete it but then chatgpt started to suggest actual genres like "Fantasy" when I was sending random nonsense like "fgsaasasadad" to it. So I decided to keep "unknown" option as last resort so to say.  And when I tested it worked fine ( but you never can be totally sure with chatgpt looks like )

andanf-e (approved):

NazarKovtun1 (approved):

PR #303, opened 2023-09-21
wr-287: Add ability to share articles with the registered users

drMangust:
I've done everything, but I don't like one thing (maybe I'm worried for nothing)

- I'm in the SharedArticleRoute wrapper, if there is a user, I do a token dispatch to the server and get the id
- then it doesn't matter what page we get to, whether on ArticlePage or SharedArticlePage, we again need to go to the server for page data (on ArticlePage, for comments)


frontend/src/libs/components/shared-article-route/shared-article-route.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const dispatch = useAppDispatch();
+  const { user, articleId, articleIdByTokenDataStatus } = useAppSelector(
+    ({ auth, articles: { articleIdByToken, articleIdByTokenDataStatus } }) => ({
+      user: auth.user,
+      articleId: articleIdByToken,
+      articleIdByTokenDataStatus: articleIdByTokenDataStatus,
+    }),
+  );
+  const { token } = useParams();
+  const hasUser = Boolean(user);
```
IrynaK-a, line 31: Could we just use the user in the next condition without creating a new variable?

drMangust: I left it as it is done in ProtectedRoute and PublicRoute. If I had chosen, I wouldn't have created it

frontend/src/libs/components/shared-article-route/shared-article-route.tsx
```diff
@@ -0,0 +1,60 @@
...
+      articleIdByTokenDataStatus === DataStatus.REJECTED
+    );
+
+    if (!articleId && !isLoading) {
+      return <Navigate to={redirectPath} />;
+    }
+
+    return (
+      <Loader isLoading={isLoading} hasOverlay type="circular">
+        <Navigate to={`${AppRoute.ARTICLES}/${articleId}`} replace />
```
andanf-e:
```suggestion
        <Navigate
          to={configureString(AppRoute.ARTICLES_$ID, { id: articleId }) as typeof AppRoute.ARTICLES_$ID;}
          replace
        />
```

drMangust: [87f0669](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/303/commits/87f066940eac038f89dfc05478eebe3dfc78fe5d)

andanf-e (approved):

mhrechkin (approved):

NazarKovtun1 (approved):

PR #359, opened 2023-09-25
wr-349: Show notification about subscribing/unsubscribing

kovAnya:


kovAnya:
I didn’t move notification messages to enum – so as they are straight in code in other files. Maybe I should create a separate NotificationMessage enum and move all notification messages there?

andanf-e:
> I didn’t move notification messages to enum – so as they are straight in code in other files. Maybe I should create a separate NotificationMessage enum and move all notification messages there?

That's okay with notification local messages, but I'm not sure about verbiage for follow notifications - maybe let's ask QAs to find out better verbiage 🤔

andanf-e (approved):

PR #360, opened 2023-09-25
wr-297: Profile page for mobile/narrow screens

hnoievets:
- [x]  Arrange Achievements, Your Writing Activity, Your Latest Articles, and Genres Stats sections in vertical order for mobile/narrow screens.
- [x]  When a user clicks on a section name, the complete information should be displayed.
- [x]  When a user clicks on an already opened section name, the information should be hidden again.
- [x]  In the "Your Writing Activity" section the most recent data should be displayed first, and a horizontal scroll bar should be displayed for data navigation.

Mobile:
image
image
image
image
image

Tablet:
image
image
image
image

Laptop(1024px):
image


frontend/src/libs/components/spoiler/styles.module.scss
```diff
@@ -0,0 +1,44 @@
...
+  max-height: 60px;
+  overflow: hidden;
+  transition: all 0.5s linear;
+}
+
+.spoiler[open] {
+  max-height: 500px;
+}
+
+summary {
```
andanf-e, line 12: Let's add class selector for summary to style it

andanf-e (commented):
And on the screenshots - there is no gap between sections on mobile screen, let's add it

PR #361, opened 2023-09-26
wr-314: Add sorting to the filters, redesign the Follow button, remove the dropdown menu for articles by other users

AnastasiaR2:
- [x] The dropdown menu icon should be displayed only for articles authored by the user.
- [x] Authors in the Author filter’s dropdown list should be sorted in alphabetical order.
- [x] Genres in the Genre filter’s dropdown list should be sorted alphabetically.
- [x] The user’s avatar should not change its size during profile editing.
- [x] Redesign the Follow button and add some padding.

image
image
image


backend/src/packages/users/user.repository.ts
```diff
@@ -163,12 +164,10 @@ class UserRepository implements IRepository {
...
 
   public async getAllAuthors(): Promise<UserDetailsModel[] | null> {
-    const authors = await UserDetailsModel.query().distinctOn('userId')
-      .whereRaw(`
-      EXISTS(SELECT 1
-      FROM articles,user_details
-      WHERE articles.user_id = user_details.user_id)
-    `);
+    const authors = await UserDetailsModel.query()
+      .from(getAuthorsSubquery())
```
andanf-e: Let's keep this query in repository - as from helper it is used only single time here

backend/src/packages/users/user.repository.ts
```diff
@@ -163,12 +163,24 @@ class UserRepository implements IRepository {
...
-      EXISTS(SELECT 1
-      FROM articles,user_details
-      WHERE articles.user_id = user_details.user_id)
-    `);
+    const authors = await UserDetailsModel.query()
+      .from((builder) => {
+        void builder
+          .distinctOn('userId')
+          .select('userId', 'firstName', 'lastName')
+          .from('user_details')
```
andanf-e, line 171: Are you sure we need to have separated .from statement as it's already in UserDetailsModel

AnastasiaR2: If I remove separated .from and leave the code like so
```
const authors = await UserDetailsModel.query()
      .distinctOn('userId')
      .select('userId', 'firstName', 'lastName')
      .from('user_details')
      .whereRaw(
        `
        EXISTS(SELECT 1
        FROM articles,user_details
        WHERE articles.user_id = user_details.user_id)
      `,
      )
      .orderBy('userId')
      .as('subquery')
      .orderBy('firstName')
      .orderBy('lastName');
```
the sorting doesn't work. Did I miss something?

andanf-e: Not sure that aliasing is needed, and I believe at first you should order by firstname\lastname,, and only after that with userId

andanf-e: And let's have sorting on authors with `orderBy('CONCAT(first_name, ' ', last_name)')` if it's possible

AnastasiaR2: As I have checked, the expressions specified in the DISTINCT ON clause must also appear in the ORDER BY clause in the same order.
So, it's not possible to order by firstname\lastname at first. That's why, I used a subquery here.

andanf-e: Okay

andanf-e (dismissed):

backend/src/packages/users/user.repository.ts
```diff
@@ -163,12 +163,24 @@ class UserRepository implements IRepository {
...
+          .whereRaw(
+            `
+        EXISTS(SELECT 1
+        FROM articles
+        WHERE articles.user_id = user_details.user_id)
```
AnastasiaR2, line 176: I noticed that all users are displayed as authors, even if no articles are written by the user. Updated query to resolve this.

andanf-e (approved):

PR #362, opened 2023-09-26
wr-297: Profile page for mobile/narrow screens

andanf-e:
Mobile:
image
image
image
image
image

Tablet:
image
image
image
image
image

Laptop(1024px):
image


andanf-e (commented):
[PR](https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/360) was closed due to branch renaming, so move the screenshots from closed one please
And comments remain the same

frontend/src/libs/components/header/styles.module.scss
```diff
@@ -88,6 +88,11 @@ img.avatar:hover {
     display: none;
   }
 
+  .header {
+    height: 80px;
```
andanf-e: Not sure about overwriting header height this way - as above it's defined through variable - won't it impact other places dependent on header height?

PR #397, opened 2023-09-27
wr-355: Change buttons style in the achievements, move the blocks header, add placeholder

IrynaK-a:
Genres Stats overlaping was fixed in wr-383 by @hnoievets

image

image


frontend/src/libs/components/button/button.tsx
```diff
@@ -5,7 +5,7 @@ import { type ReactMouseEvent, type ValueOf } from '~/libs/types/types.js';
 import { Loader } from '../components.js';
 import styles from './styles.module.scss';
 
-type ButtonVariant = 'primary' | 'outlined' | 'text';
+type ButtonVariant = 'primary' | 'greenOutlined' | 'whiteOutlined' | 'text';
```
andanf-e: Let's separate it to `color` prop and keep outlined variant as it was

IrynaK-a: Isn't it better to have two variants of that button, in accordance with our design, I believe, we have only two outlined buttons.
Do we need to get the opportunity to set outlines in different colors?)

andanf-e: I thought that mean that main color of button would be green\white

frontend/src/libs/components/button/button.tsx
```diff
@@ -48,6 +52,11 @@ const Button: React.FC<Properties> = ({
     small: styles.buttonSmall,
   };
 
+  const outlineColorClassNameMapper: Record<ButtonOutlineColor, string> = {
+    green: styles.greenOutline,
+    white: styles.whiteOutline,
```
andanf-e: Could it be styles.green\white? as it's not only about border, but also text color

frontend/src/pages/profile/components/user-latest-articles/user-latest-articles.tsx
```diff
@@ -18,7 +19,13 @@ type Properties = {
...
 const UserLatestArticles: React.FC<Properties> = ({ className }) => {
   const dispatch = useAppDispatch();
-  const { articles } = useAppSelector(({ articles }) => articles);
+  const { articles, articlesStatus } = useAppSelector(({ articles }) => ({
+    articles: articles.articles,
+    articlesStatus: articles.dataStatus,
+  }));
+
+  const isLoadingArticles = articlesStatus === DataStatus.PENDING;
+  const isArticles = Boolean(articles.length) || isLoadingArticles;
```
andanf-e: Not sure about this check and condition

frontend/src/pages/profile/components/user-articles-genres-stats/user-articles-genres-stats.tsx
```diff
@@ -61,63 +61,66 @@ const UserArticlesGenresStats: React.FC<Properties> = ({ className }) => {
...
-                  dataKey="count"
-                  labelLine={false}
-                  data={chartData}
-                  innerRadius={GenresChartConfig.INNER_RADIUS}
-                  outerRadius={GenresChartConfig.OUTER_RADIUS}
+      <div className={getValidClassNames(styles.wrapper, className)}>
+        <div className={styles.chartWrapper}>
+          {userArticlesGenresStatsStatus === DataStatus.PENDING && (
+            <Loader isLoading type="dots" />
+          )}
```
andanf-e: Can't we wrap up a chart with it?

frontend/src/libs/components/button/button.tsx
```diff
@@ -48,6 +52,11 @@ const Button: React.FC<Properties> = ({
     small: styles.buttonSmall,
   };
 
+  const outlineColorClassNameMapper: Record<ButtonOutlineColor, string> = {
+    green: styles.green,
+    white: styles.white,
+  };
```
andanf-e: As of now keys are directly matching the classes - so we can pass them directly I guess

frontend/src/pages/profile/components/user-articles-genres-stats/user-articles-genres-stats.tsx
IrynaK-a: Sorry, I don't get it. You mean can we wrap up a chart with DataStatus.PENDING condition?

andanf-e: Yep, with Loader

frontend/src/pages/profile/components/user-latest-articles/user-latest-articles.tsx
```diff
@@ -18,7 +19,11 @@ type Properties = {
 
 const UserLatestArticles: React.FC<Properties> = ({ className }) => {
   const dispatch = useAppDispatch();
-  const { articles } = useAppSelector(({ articles }) => articles);
+  const { articles } = useAppSelector(({ articles }) => ({
+    articles: articles.articles,
+  }));
+
+  const isArticles = Boolean(articles.length);
```
andanf-e: Maybe better would be to name it `hasArticles`

frontend/src/pages/profile/profile-page.tsx
```diff
@@ -20,10 +20,15 @@ import styles from './styles.module.scss';
...
-  }));
+  const { user, userActivities, articles } = useAppSelector(
+    ({ auth, users, articles }) => ({
+      user: auth.user as UserAuthResponseDto,
+      userActivities: users.userActivities,
+      articles: articles.articles,
+    }),
+  );
+
+  const isArticles = Boolean(articles.length);
```
andanf-e: Same about `hasArticles`

frontend/src/pages/profile/components/user-articles-genres-stats/user-articles-genres-stats.tsx
IrynaK-a: But we add a Loader when stats are loading and then render the data (on fulfilled. Isn't it write?

andanf-e: And the Loader component on isLoading state replaces wrapped component with loader - I believe this would work the same way

andanf-e (dismissed):

andanf-e (approved):

PR #399, opened 2023-09-27
wr-383: Add placeholder to genres stats when there is no written articles

hnoievets:
- [x]  add placeholder
- [x]  disable filter when there is no articles

image

image

And fixed Genres Stats overlapping:
image

image

image


andanf-e (commented):
image
Maybe it would be better to have something like that as a placeholder for chart
Please check if it could be directly achieved with recharts

frontend/src/pages/profile/components/user-articles-genres-stats/libs/constants/constants.ts
```diff
@@ -11,10 +11,23 @@ const GENRES_CHART_COLORS = [
   '#EAFCFF',
   '#E6F4F1',
 ];
+const GENRES_CHART_PLACEHOLDER_DATA = [{ lable: 'No data', count: 1 }];
+const GENRES_CHART_PLACEHOLDER_COLOR = '#f0eded';
+const GENRES_CHART_PLACEHOLDER_FONT_SIZE = 18;
+const GENRES_CHART_PLACEHOLDER_LABEL_COLOR = 'gray';
+const GENRES_CHART_PLACEHOLDER_LABEL_POSITION = 'center';
```
andanf-e: Let's group it into a `GenresChartPlaceholder` enum

andanf-e (approved):

PR #409, opened 2023-09-27
wr-405: Fix loader inside share popover

kovAnya:


kovAnya:
image

frontend/src/libs/components/share-popover/share-group.tsx
```diff
@@ -49,11 +49,7 @@ const ShareGroup: FC<Properties> = ({ articleId: id, articleTitle }) => {
   }
 
   return (
-    <Loader
-      isLoading={isLoading}
-      type="circular"
-      overlayClassName={styles.loader}
-    >
+    <Loader isLoading={isLoading} type="dots" overlayClassName={styles.loader}>
```
andanf-e, line 52: Can we place Loader only on Share Facebook button?

drMangust: I've got you confused. It won't work that way now. The flow now is such that there is only one link for everything. That is, before clicking on any item, we need to get a link from the backend

andanf-e: 🤔 
Got it

andanf-e (approved):

frontend/src/libs/components/share-popover/share-group.tsx
drMangust: You can make menu items inverse. Or write something funny, like we're preparing a link. But if Anya can find a way to solve this issue, it will be cool. And once again, I chose this option and removed the previously created functionality due to the fact that when we open the modal, we are already sending a request for the backend, and doing it again to click on the copy button would not be very good

PR #422, opened 2023-09-28
wr-414: Fix the displaying article text with a bigger font size

IrynaK-a:
image
image

### DIFF
--- a/frontend/src/pages/articles/components/article-card/article-card.tsx
+++ b/frontend/src/pages/articles/components/article-card/article-card.tsx
@@ -21,6 +21,7 @@ import {
   getReactionConvertedToBoolean,
   getReactionsInfo,
   getValidClassNames,
+  modifyFontSize,
   sanitizeHtml,
 } from '~/libs/helpers/helpers.js';
 import {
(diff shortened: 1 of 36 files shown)


frontend/src/pages/articles/components/article-card/article-card.tsx
```diff
@@ -191,7 +192,10 @@ const ArticleCard: React.FC<Properties> = ({
           <h4 className={styles.title}>{title}</h4>
           <article
             className={getValidClassNames(styles.text, 'text-overflow')}
-            dangerouslySetInnerHTML={{ __html: sanitizeHtml(text) }}
+            style={{ fontSize: 14 }}
+            dangerouslySetInnerHTML={{
+              __html: sanitizeHtml(modifyFontSize(text, '14px')),
+            }}
```
andanf-e: Can't we just make it with classes styling?

IrynaK-a: Should I use !important?
because we receive the `text` with inline style.

andanf-e: If with important we'd need just to provide styling (without replacing anything in text) - then I believe you can

andanf-e (approved):

PR #430, opened 2023-09-28
wr-415: Fix the redirection to the comment block on Feed page

hnoievets:


frontend/src/pages/article/article-page.tsx
```diff
@@ -81,13 +81,14 @@ const ArticlePage: React.FC = () => {
   }, [location]);
 
   useEffect(() => {
-    void dispatch(articleActions.getArticle(Number(id)));
+    void dispatch(articleActions.getArticle(Number(id))).then(scrollToComments);
```
andanf-e, line 84: Won't it behave the way that on article page opening there would always be scrolling to comments?

hnoievets: No, it won't. It scrolls only if we click on the comments icon, because only in this case `hash` has a value

andanf-e (approved):

PR #442, opened 2023-09-29
wr-435: Fix autozooming for narrow screens

kovAnya:


kovAnya:
I've set font to 16px for mobile screens and checked for side-effects by ngrok

andanf-e (approved):

PR #451, opened 2023-09-29
wr-419: Display notification about New articles written by followed authors on all pages

PablloZz:
image
image
image


backend/src/db/migrations/20230928125310_update_user_achievements_description.ts
andanf-e, line 1: Don't change existing migrations

backend/src/db/migrations/20230929065718_update_user_achievements_description.ts
```diff
@@ -0,0 +1,232 @@
...
+} as const;
+
+async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TABLE_NAME, (table) => {
+    table.dropColumn(ColumnName.DESCRIPTION);
+  });
+
+  await knex.schema.alterTable(TABLE_NAME, (table) => {
+    table.jsonb(ColumnName.DESCRIPTION);
+  });
```
andanf-e: I guess to update description text you don't need to recreate column
```suggestion

```

backend/src/db/migrations/20230929065718_update_user_achievements_description.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  await knex.schema.alterTable(TABLE_NAME, (table) => {
+    table.text(ColumnName.DESCRIPTION);
+  });
+
+  for (const achievement of ACHIEVEMENTS) {
+    await knex(TABLE_NAME)
+      .where(ColumnName.KEY, achievement.key)
+      .update({ [ColumnName.DESCRIPTION]: achievement.description });
+  }
```
andanf-e: Same on rollback - +as you're correcting description text -- you can leave rollback as empty

backend/src/db/migrations/20230928125310_update_user_achievements_description.ts
PablloZz: Weird, I seem to have deleted the old migration

frontend/src/libs/hooks/use-articles-feed-room/use-articles-feed-room.hook.ts
artemmatiushenko1, line 1: Maybe it will be worth creating a separate hook for handling global notifications because here we also listen to the `NEW_REACTION` event and dispatch `articleActions.addReactionToArticlesFeed(reaction)` action, so at the time we receive this event we might be on a different page and some unnecessary logic might be executed (potentially causing some unexpected side effects)

We can leave logic that writes some data to the store in this hook, but move the notifications-related code to a separate one that will be used here `frontend/src/libs/components/app/app.tsx`. What do you think?

frontend/src/libs/hooks/use-articles-feed-room/use-articles-feed-room.hook.ts
```diff
@@ -40,8 +47,24 @@ const useArticlesFeedRoom = (): void => {
     articlesSocket?.on(
       NEW_ARTICLE,
       (newArticle: ArticleSocketEventPayload[typeof NEW_ARTICLE]) => {
-        if (location.pathname === AppRoute.ARTICLES) {
-          void dispatch(articleActions.addArticle(newArticle));
+        if (user?.id === newArticle.socketUserId) {
```
artemmatiushenko1, line 50: I am not sure why we need to send `socketUserId` in the event's payload, can't it be 
`if (user?.id === newArticle.article.id) return` ? Maybe I misunderstood something 😅

frontend/src/libs/hooks/use-articles-feed-room/use-articles-feed-room.hook.ts
artemmatiushenko1, line 1: Maybe you should cooperate with @canterbery, your tasks seem to be related https://github.com/BinaryStudioAcademy/bsa-2023-writorium/pull/458

frontend/src/libs/hooks/use-articles-feed-room/use-articles-feed-room.hook.ts
PablloZz: In this case we only check the author of the article, but on the server we loop through all the sockets in the namespace and send to each user info about each socket on every loop cycle (for example if we have 5 sockets each user will get 5 payloads), so if only one user follows the author, every user will be notified of this publication due this one payload from that user.

PablloZz: Well, maybe it could have been done better, but I was in a bit of a rush to meet the deadline)

andanf-e (approved):

